PR #9, opened 2014-11-14
Numbered Captions and Total Number of Frames

dudektria:
I made captions numbered and the total number of frames show up next to the current frame number.
I thought it might be useful.


dudektria:
Sorry for the late replay (it _is_ late indeed!). As you said, the total number of slides seems obsolete when used with the progress bar, so I will not bother adding it anymore :)
The captions, on the other hand, seem more useful. I will make a new pull request, OK?

matze:
Yeah sure, go ahead!

PR #12, opened 2014-12-10
* beamerthemem.sty(frametitle): \protect the frametitle inside MakeLowercase

daimrod:
Without the \protect, we cannot use \cite, \ref or similar stuff in frametitle.


matze:
I reverted this change in favor of a package option, I hope you don't mind. `\protect` nullifies `\MakeLowercase` but if you want to have this, you can now add `protectframetitles` to the `\usetheme` command.

Issue #8, opened 2014-11-12
Make theme compatible with stock color themes


matze:
Especially the background is not set correctly.

rchurchley:
I have tested cc24402 with the default beamer colour themes: `albatross`, `beaver`, `beetle`, `crane`, `default`, `dolphin`, `dove`, `fly`, `lily`, `orchid`, `rose`, `seagull`, `seahorse`, `sidebartab`, `structure`, `whale`, and `wolverine`. It works essentially as expected, with the following quirks:

- Several themes do not define some colours, expecting to inherit them from `default`; for example, they may intend to have red `alerted text` as per the default but do not specify this. When used with our theme, these styles will fall back to `mtheme`'s colours instead of the defaults. We may want to provide some option for `beamercolorthememetropolis` to *not* set the base colours `normal text`, `alerted text`, and `example text` to allow users to let users revert to the default colour scheme.
- The `\plain{...}` command does not work with the `fly` theme. This is because `fly` has an annoying definition of `palette primary` which specifically *unsets* the background. I don't think there's a non-hacky way of fixing this while maintaining the expected behaviour for the other themes.

benjamin-weiss:
> I have tested cc24402 with the default beamer colour themes.

Wow. Sounds like a lot of boring work.

> We may want to provide some option for `beamercolorthememetropolis`

Maybe even go a step further and provide an option not to load the color theme at all. And instead only define the new beamer colors we use?

    \if@nocolortheme
      \setbeamercolor{title separator}{use=progress bar, parent=progress bar}
      \setbeamercolor{progress bar}{%
        use=alerted text,
        fg=alerted text.fg,
        bg=normal text.bg!50!fg
      }
    \else
      \usecolortheme{metropolis}
    \fi

If I get it right, this should also fix the problem and an option like this should be easier to understand than `notSetBaseColors`.

> The `\plain{...}` command does not work with the `fly` theme.

It appears to me that this is a bug of the `fly` theme. I wouldn't really care about that. Lets just keep it in mind and add it to the documentation as a known issue.

rchurchley:
> Maybe even go a step further and provide an option not to load the color theme at all. And instead only define the new beamer colors we use?

After splitting the theme into its inner, outer, font, and colour components, it is now possible to take advantage of the structural parts of the theme without loading the colour theme. However, Beamer colors for (e.g.) `progress bar` are not defined if the color theme is not loaded. Should we provide defaults in the inner and outer themes?

[...]

PR #21, opened 2015-01-23
Use palette colors instead of hardcoded names

jdahm:
Applied this for a presentation recently with a different color theme and really liked it! I made some changes to use palette colors defined by the chosen color theme instead of the hardcoded names. I used some care and tried to ensure that the changes it makes do not affect the look of the demo presentation.

The second commit fills in a few more color definitions, specific to blocks, which were omitted from the theme. Users can then add \useinnertheme{rounded} or \setbeamertemplate{blocks}[rounded] and it should give nicely formatted blocks.

This is my first pull request, and it seems that I accidentally slipped in a change to the .gitignore file. Sorry, I don't know how to remove it from the pull request...


matze:
Very, very nice! I had this idea lingering around for a while, also in order to fix #8. But one thing, could you split off the block colors in its own color theme? Leaving out the block colors was a deliberate decision. I (yes, personally) find those heavy stock blocks extremely distracting but like them for structuring content.

jdahm:
I added an option to both the main as well as the color theme. Now a user can load the blocks with `\usetheme[blockbg]{m}`. Also, I realized the block colors weren't very good, so I tweaked them to be more soothing.

matze:
> I tweaked them to be more soothing.

:smile:

Thanks for the contribution.

PR #23, opened 2015-01-23
Simple fix to ensure colour of footnote mark matches the surrounding environment.

kdungs:
Awesome theme!

When using `\plain` the footnote marks were a bit too dark. This should fix it.

Before:
screen shot 2015-01-23 at 1 18 37 pm

After: 
screen shot 2015-01-23 at 1 45 09 pm


matze:
Thanks!

PR #30, opened 2015-02-20
Added optional support for title graphics

Leonidas-from-XIV:
Sometimes it is useful, especially when waiting to present, to have something visual people can look at while waiting for the presentation to begin.

Taken from the themes shipped with Beamer.

### DIFF
--- a/beamerthemem.sty
+++ b/beamerthemem.sty
@@ -60,6 +60,11 @@
 {
   \begin{minipage}[b][\paperheight]{\textwidth}
     \vfill
+    \ifx\inserttitlegraphic\@empty%
+    \else%
+    {\usebeamercolor[fg]{titlegraphic}\inserttitlegraphic\par}%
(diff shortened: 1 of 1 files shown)


matze:
Looks a bit funny to me but if it works for you …

Thanks for the contribution.

benjamin-weiss:
I also think the positioning is a bit odd. Here is a different proposal:

https://github.com/benjamin-weiss/mtheme/commit/cd8635b2113c012e23bc8361f01dabb4937e4eec

Although I really don't like my implementation (this kind of positioning must be possible in a much easier way), I think you will get the idea from this.

Probably for a lot of logos a position on the upper right side of the title page is the way to go. Yet I would not advise to add a option or make it the standard behavior, as you can easily set it inside your main document, using:

```
\titlegraphic{\raggedleft\includegraphics[width=0.3\linewidth]{logo}}
```

So, what do you think?

PR #32, opened 2015-02-22
Title graphics / section title line break

benjamin-weiss:


matze:
Many thanks for the changes and fixes. I couldn't chime in earlier about the title graphics issue but I agree with you that this makes more sense (albeit it's not really nice to look at).

By the way, thanks also for the HSRM theme which was a great initial example and motivation for this theme.

benjamin-weiss:
Nothing to thank for. I really like what you did with your theme. Your more minimalistic approach is even better than my HSRM theme. And overall it is much more universal than my theme.

What do you think about the ```\textsc{\MakeLowercase{…}}``` problem. Why did you do that in the first place? (look at the comment @ https://github.com/benjamin-weiss/mtheme/commit/629c117f745069862d82838385639629a8b689d4)

matze:
Using small caps of lower-cased letters looks a whole lot different than just uppercasing them. So I use that purely for aesthetic reasons.

benjamin-weiss:
Ok, I see. When there is a real *small caps* font style with the Fira font: why not just use *small caps* and drop the ```\MakeLowercase``` or ```\MakeUppercase```? Have you tested that? How does it look like?

FYI, I just used the ```\MakeUppercase``` in my theme, because it was part of the Corporate Design. I don't really think this is good typesetting.

matze:
> why not just use small caps and drop the \MakeLowercase or \MakeUppercase?

Fira Sans has small caps for lower and uppercase letters which is maybe okay for single-world titles (see image) but looks a bit wonky for longer "camel-cased" sentences.

comparison

y1lan:
I would like to know that how can I use title graphics?

Issue #38, opened 2015-04-09
Blocks


chraibi:
Like your theme.  I have a slight problem with the blocks. Why they are showing now "visible" color in the header? (In the demo.pdf)

matze:
Could you elaborate a bit more? Neither the provided demo not a recompiled one has any colors. This is a recent feature that must be enabled explicitly with the `blockbg` option.

chraibi:
how can I set the option `blockbg`?

jdahm:
Hi chraibi,

I added this feature recently quite quickly and I didn't fully test it. Thanks for the input. I don't have a lot of time right now to work on this, but the relevant section is lines 60-70 of beamercolorthememetropolis.sty. From a quick scan it looks like I forgot to specify bg= in a few lines
```tex
\setbeamercolor{block title alerted}
\setbeamercolor{block title example}
```
and
```tex
\setbeamercolor{block title alerted}
\setbeamercolor{block title example}
```
Adding a pull request now. Please test this.

matze:
Fixed with #41.

PR #41, opened 2015-04-17
Add bg color to alerted and example blocks.

jdahm:
Addresses #38. Please test.

### DIFF
--- a/beamercolorthememetropolis.sty
+++ b/beamercolorthememetropolis.sty
@@ -60,12 +60,12 @@
 \if@beamer@metropolis@blockbg
 
 \setbeamercolor{block title}{use=palette primary,parent=palette primary,bg=palette primary.bg!80!fg}
-\setbeamercolor{block title alerted}{use=alerted text,parent=palette primary,fg=alerted text.fg,bg=palette primary.bg!80!fg}
-\setbeamercolor{block title example}{use=example text,parent=palette primary,fg=example text.fg,bg=palette primary.bg!80!fg}
+\setbeamercolor{block title alerted}{use={palette primary,alerted text},parent=palette primary,fg=alerted text.fg,bg=palette primary.bg!80!fg}
+\setbeamercolor{block title example}{use={palette primary,example text},parent=palette primary,fg=example text.fg,bg=palette primary.bg!80!fg}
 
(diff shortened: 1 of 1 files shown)


matze:
Something's off with the height of both the block title and the block itself.

jdahm:
@chraibi Thanks for testing. Just had a chance to put together a test myself and incorporated your changes. I've cleaned up the syntax a bit and increased the contrast between the block title and body backgrounds, which is a bit lacking in your example. Adding the commit now.

If you would like to revert to what you had, just add in the preamble
```tex
\usetheme[blockbg]{m}
\useinnertheme{rectangles}

\makeatletter
\setbeamercolor{block title}{use=palette primary,parent=palette primary,bg=palette primary.bg!80!fg}                             
\setbeamercolor{block title alerted}{use=alerted text,parent=palette primary,fg=alerted text.fg, bg=alerted text.bg!75!fg}       
\setbeamercolor{block title example}{use=example text,parent=palette primary,fg=example text.fg,bg=example text.bg!75!fg}        

\setbeamercolor{block body}{parent=normal text,use=block title,bg=block title.bg!80}                                             
\setbeamercolor{block body alerted}{parent=normal text,use=block title alerted,bg=block title alerted.bg,bg=alerted              
text.bg!80!fg}                                                                                                                   
\setbeamercolor{block body example}{parent=normal text,use=block title example,bg=block title example.bg,bg=example              
text.bg!80!fg}       
\makeatother
```

@matze I noticed that when I used this theme as well (now trying to transition to reveal.js, but that's unrelated here). The cause is that beamer seems to be defaulting to something like the `rectangles` inner theme. This is either since mtheme sets this (which it doesn't) or this is a beamer default. If you instead use something like `\useinnertheme{rounded}` the spacing is much better. It would be nice if someone added better looking default spacing settings to mtheme itself.

matze:
@jdahm: so it's ready for merge?

jdahm:
Yes, it is. Not quite the colors shown above (I made the background lighter) but we'll never find a setting that pleases everyone. It can always be customized.

chraibi:
@jdahm yes that looks better. Thanks! 
Apropos your remark concerning ```\useinnertheme{rounded}``` that makes the spacing much better indeed. Only the master slide gets totally changed, which is not nice.

PR #45, opened 2015-04-24
Add three options: usesmallcapitalsintitle, usesectionslide and nototals...

Schwefelsaeure:
I have added **three options** to let the author decide how the beamer presentation should look like instead of forcing a specific behavior:

- **usesmallcapitalsintitle**: print frame titles in small capitals. I made this optional because small capitals are harder to read than alternating cases.
- **usesectionslide**: introduce a new section by an own slide. For short presentations with only a few slides, it is overkill to introduce each section with an own slide.
- **nototalslides**: print "#current/#total" slides per default. This indicates the progress and introduces almost no visual noise. Furthermore, I have used **\insertpagenumber** instead of **\insertframenumber** because each slide should have an unique "identifier" where questions can relate to. When using overlays, **\insertframenumber** uses the same frame number for each overlay.

See the diff for all code changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,16 @@ make content more centered on the frame. If using more content per
 slide, this can be turned off at the package-level by passing the
 `nooffset` option.
 
+With option `nosectionslide`, no dedicated slide is produced when a new section
+starts. By default when using the `\section` command, a slide is created with
+just the title on it.
+
+The `nosmallcapitals` option supresses the usage of small capitals on
+title page, frame titles and dedicated section slides.
+
+Option `usetotalslideindicator` creates slide numbering in lower right corner
+in following format: #current/#total. By default, just current page number is
+printed.
 
 #### Commands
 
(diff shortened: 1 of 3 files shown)


matze:
I think these are valuable additions to the theme however in my opinion these should not be defaults. I'd rather opt in to use non-small caps, section slides and totals. Here are some counter arguments:

> I made this optional because small capitals are harder to read than alternating cases.

This is a very subjective topic and personally I doubt this is generally true.

> For short presentations with only a few slides, it is overkill to introduce each section with an own slide.

Then, why even bother using sections? Apart from the PDF TOC there is no way to tell that they exist.

> This indicates the progress and introduces almost no visual noise.

In my opinion it does add enough visual noise. The `usetitleprogressbar` allows progress indication in a subtle way without reminding each and every slide how many slides there are in total.

P.S.: You may also document these options in the README.

Schwefelsaeure:
> I think these are valuable additions to the theme however in my opinion these should not be defaults.

Fully aggree on that. Now, all these options are disabled by default. Therefore, they must be enabled explicitly. Furthermore, I renamed the options to make this change visible:

- usesmallcapitalsintitle -> nosmallcapitals
- usesectionslide -> nosectionslide
- nototalslides -> usetotalslideindicator

> P.S.: You may also document these options in the README.

I added a description for all three options in README.

PS: Thanks four your feedback. :)

matze:
Thanks for your contribution!

PR #47, opened 2015-04-28
Reduce dependencies to help new users

### LINKED ISSUE #7: Version requirements
Fedora 20 does ship `Fira {Sans,Mono}` as `Fira {Sans,Mono} OT` and also lacks the `ExtraLight` variant, causing xelatex to thrown upon.

0.3.x is the minimum I guess, at least I did not see the `ExtraLight` in the 0.2.x tarball. About the OT not sure if this is fedora specific or also a change between 0.2.x and 0.3.x

rchurchley:
I'm suggesting a couple changes to make it easier for new users to get started with the `mtheme`:

- remove `minted` and the dependency on Pygments
- move `\usecolortheme` and `\usefonttheme` to `demo.tex`

The first change makes a different tradeoff between the beauty and usability of the demo slides. Using `minted` is great for making the slides a beautiful demonstration of what the `mtheme` can do. However, many new users compile demo slides as a test to see whether the theme works in their environment. They may give up on the `mtheme` as being too complicated to set up when they conclude that `--shell-escape` and `Pygments` are required to compile the demo slides — even though they're not needed to actually use the `mtheme`.

The second change is pretty minor but potentially useful. It simply moves the `\usecolortheme{metropolis}` and `\usefonttheme{metropolis}` declarations from `beamerthemem` to the `.tex` file. This resolves #7 and helps with #8, making it a bit easier to work around font dependencies and to use a custom colour theme per presentation.


matze:
I agree with your first point of removing minted for the sake of easier demo builds. However,

> The second change is pretty minor […]

doesn't sound quite right to me. The point of a (named) theme is to set some defaults concerning colors, shapes etc. Just have a look at the default themes distributed with Beamer.

Using a custom color theme (should) not be a problem by setting it after loading the theme. To improve this and allow for real customization we should maybe think about splitting things up even more into inner and outer themes, however – and this is my personal opinion – this theme was always intended to be "opionated".

rchurchley:
Ah, you're right — I was mistaken about the structure of beamer themes, and of course colours can be overridden later. I agree that the colour theme declaration should stay in `beamerthemem.sty`.

Including the font theme is a bit different technical situation: including it by default adds dependencies on XeLaTeX and a particular name and version of the Fira font families. I like the opinionated nature of the `mtheme`, so customization is not my concern here; I'd just like an easy way to draft slides on my Fira-less work computer before finalizing them with the correct fonts at home. Moving the font theme declaration to the `.tex` file was the easy fix, but we could add an option to enable or disable the font theme declaration in `beamerthemem.sty`. I'm fine with either choice, though I'm aware that the list of package options is growing and would not want their number to diminish the theme's strong personality.

matze:
Alternatively, we could also add a Fira-less base theme which loads everything except for the font declaration and a full version which includes the base theme + the fonts.

rchurchley:
Good point. In any case, the font theme issue is probably worth considering in a future update. I've changed the pull request to only address the `minted` issue.

matze:
Alright, you are welcome to open more pull requests concerning the font issue :smile:.

PR #48, opened 2015-04-30
Tidy title slide code

rchurchley:
This pull request consists entirely of bugfixes and code formatting improvements, and should not affect presentation output except in corner cases.

- Fixes a bug causing unwanted space on the title slide when given an empty `\institute`. (Typo.)
- Fixes a bug causing unwanted space on the title slide when given an empty `\author`. (Apparently `\insertauthor` [works differently than the other title page elements](http://tex.stackexchange.com/questions/241306/).)
- Fixes the root cause of extra space following the title graphic, which previously had to be worked around with a negative vspace.
- Provides a more elegant solution for positioning title graphics.
- Makes the code more git-friendly by putting each formatting command on its own line.

### DIFF
--- a/beamerthemem.sty
+++ b/beamerthemem.sty
@@ -79,7 +79,7 @@
 		\setbox0=\vbox{\inserttitlegraphic}%
 		\logoheight=\ht0 \advance\logoheight by \dp0 %
 		\vspace*{-\logoheight}%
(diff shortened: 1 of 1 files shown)


matze:
Very nice changes, thank you!

Issue #46, opened 2015-04-24
Progress bar in footer?


foreachthing:
How can I put the progress bar in the footer, without inserting the section head?
Thanks!

(sorry, if such questions don't belong here - didn't know where to go/put)

matze:
Hmm, good question. I haven't tried that yet but do you think that makes sense? Where would that bar be located? Right on the bottom of the frame?

foreachthing:
Yeah ... something like that (see file attached).
![beamer_progress_footer](https://cloud.githubusercontent.com/assets/10420187/7500490/52e82e4c-f42e-11e4-8d9e-e9e1c3e04d41.PNG)
Weather it makes sense or not ... I'd like to know how much longer I have to sit through the presentation! :-)

benjamin-weiss:
Have you seen the `usetitleprogressbar` option of the theme? If I understand you right, this is exactly what you want. With the only difference that the progress bar is right under the frame title instead of in the footer. And from a design point of view this even makes more sense (in my opinion).

matze:
@foreachthing: do you still see a valid reason for having the progressbar on the bottom rather than on the title bar?

rchurchley:
I ran into one possible use case last week, when I gave a diagram-heavy talk that looked best with no frame titles at all. (The progress bar is never displayed when the frame title is empty.)

It could be valid in such cases to want a progress bar somewhere outside the frame title template. From a design point of view, I'd prefer it to be snug (no margins) against the top or bottom of the slide. This should probably wait until we have key-value options, though, so we can use memorable options like

```
[progressbar=none/title/top/bottom]
```

foreachthing:
Matze: A VALID reason? No, I guess not. I just like it better - is that a reason ;-)
It's like with a program that's doing a long job: I'd like to know every percent of progress, not just 20%, 40%, 100% done.

I like rchurchley's idea! Margin-less top or bottom.

matze:
Alright, then let's wait until #90 is merged and we can start thinking of an implementation.

PR #54, opened 2015-05-21
Refactor progress bar and section page

### LINKED ISSUE #44: Can't figure out how to make Section slides dark teal.
I've run the make command from the root directory, but my Section slides have a white background and dark teal progress bar. How do I specify that the section slides should appear like in the example pdf?

rchurchley:
This pull request:
- Simplifies the code responsible for displaying the progress bar. A new `\progressbar{width}` command makes it possible to reuse the same code for section pages, the frame title, and any future progress bars (#46?)
- Reimplements the beginning-of-section frames using the recommended `section page` beamer template.
- Fixes #44 by reverting the beginning-of-section frames to `palette secondary`.


matze:
Thanks Ross, really great initiative. Now I am even able understand the code with just taking a glimpse at it ;-)

However, there are some issues that need to be resolved before merging it:

- I deliberately changed the section frames to light backgrounds because I found the dark ones too distracting. I'd rather like to have the dark backgrounds optional. So maybe the real fix should be rebuilding the demo with current color settings ;-)

- The section title is not perfectly aligned with the left position of the bar.

- The background of the progress bar is white but it should be a subdued mix of teal and orange just like the title progressbar …

- … speaking of which now has a very thin gray border which makes it look cheap.

- This is not concerning the theme but the merge process. IMHO PRs should be made from branches other than master and more importantly frequenct "Merge branch 'master'" should be avoided. Ideally you would squash your changes and rebase them on top of my master so that only the relevant commits are part of the PR.

rchurchley:
I'm still reasonably new to using git collaboratively, so I really appreciate the help on merging properly! (I think) I've squashed the changes for the current request and will PR from a different branch next time.

The revised PR:
- Changes the section colours back (I'll let you update the image in the README)
- Changes the background of all progress bars to be 50% teal
- Removes the gray border from the title progress bar (apparently that's the border of the enclosing `beamercolorbox`, which I failed to update the height of)

The problem with aligning the section title with the progress bar is a tricky one and will require more research. It appears to be fundamental to how `tikzpicture`s align with external text; you can reproduce the problem as easily as this:
```latex
\documentclass{beamer}
\usepackage{tikz}
\begin{document}
\begin{frame}
    Egad!

    \begin{tikzpicture}\draw (0, 0) -- (\textwidth, 0);\end{tikzpicture}
\end{frame}
\end{document}
```
I don't think it's a case of having an extra space somewhere, since the extent to which the title is misaligned depends on the first character of the section title. The alignment is near-perfect when the title begins with an *A* or *M*, but it's noticeably off with an *E*. The title slide actually suffers from this problem as well — the good alignment on the demo slides is a coincidence caused by the first letters of each line.

It's up to you what you want to do at this point. If you think the more understandable (and more easily extendable) source code is worth the regression on the text alignment, you could accept the pull request for the time being and we could look for a fix moving forward.

matze:
> It's up to you what you want to do at this point.

It's perfectly fine :+1:, if I have a spare minute I will try to fix the alignment issue.

> I'll let you update the image in the README.

No problem, that's actually necessary for quite some time now.

PR #56, opened 2015-05-25
mention Launchpad PPA package for mtheme and Fira

eddelbuettel:


matze:
Thanks for setting up the repository :+1:

eddelbuettel:
Well it is not a new repository -- I had it already for other R work -- but it took me a bit of fiddling to get the package right, including an assist from Norbert, Debian's core maintainer for `texlive`.

The PR I sent you points to a 0.1.0 deb. I may replace that with 0.1.1 etc in due course, so we may need to find a better way (such as just [pointing to the PPA](https://launchpad.net/~edd/+archive/ubuntu/misc/+packages).  

Lastly, I also have my local modification 'm2' in there, but mtheme and Fira and in it un modified.

Issue #33, opened 2015-03-15
All-lowercase small caps titles ugly with brackets, apostrophes, numbers 


gmbrauer:
All-lowercase small caps for titles look nice as long as they contain letters only.
It all starts looking very ugly, however, as soon as brackets, apostrophes, quotation marks, or numbers are added.
As noted before (https://github.com/benjamin-weiss/mtheme/commit/629c117f745069862d82838385639629a8b689d4), \textsc{\MakeLowercase{…}} should be replaced by \MakeUppercase{…} and the fontsize adjusted accordingly.
Using `Numbers=Lowercase` seems to solve this issue for numbers, but not for other punctuation characters. For possible solutions - which do not seem to work for Fira Sans, though - see also https://tex.stackexchange.com/questions/170293/parentheses-and-square-brackets-contextually-in-small-caps.

matze:
Thanks for the feedback, I'll look into that soon. However, I think faking small caps using smaller upper case letters is a typographically wrong solution and will definitely look awkward.

matze:
#58 fixed this issue in a quite elegant way.

Issue #20, opened 2015-01-22
"TeX capacity exceeded" when using allowframebreaks


gmbrauer:
This MWE:

````
\documentclass[titleprogressbar]{beamer}
\usetheme{m}
\usepackage{lipsum}
\begin{document}

\begin{frame}[allowframebreaks]
\frametitle{A Frame Title}
\lipsum
\end{frame}

\end{document}
````

bombs with 

````
...
Runaway argument?
! TeX capacity exceeded, sorry [main memory size=5000000].
\reserved@a #1#2->\let #2#1\reserved@a 
                                       
l.9 \end{frame}
...
````

matze:
Good catch. I don't know what to do about it right now but I will investigate.

boppreh:
I just hit this problem too.

big-r81:
I just hit this problem too.
I need [allowframbreaks] for the \printbibliography command

rchurchley:
This issue has now been fixed as of #58 and can probably be closed (independently of #59)

PR #58, opened 2015-05-28
Remove MakeLowercase

rchurchley:
Although it makes the small-caps frame titles look nice, the `mtheme`'s use of `\MakeLowercase` causes several other problems:

- The crash described in #20 is (somehow) caused when `\MakeLowercase` is called in the automatically generated extra frame titles.
- As #33 points out, `\MakeLowercase` only changes the case of letters in the English alphabet, producing undesirable results in common cases where the frame title contains numerals or non-Latin Unicode characters.
- Finally, `\MakeLowercase` sets a potentially bad trap for users. You might want to give a talk about "ᴀɴ *O(n²)* ᴀʟɢᴏʀɪᴛʜᴍ", for example, and write the obvious `\frametitle{An $O(n^2)$ algorithm}`. This appears to work as desired: the title font and small-caps shape do not apply to the math-mode part. But on closer inspection, `\MakeLowercase` *does* apply in math mode, so the title's meaning is unexpectedly distorted to "ᴀɴ *o(n²)* ᴀʟɢᴏʀɪᴛʜᴍ".

These downsides are significant enough to consider giving up the automatic lower-casing of titles: it's easy enough for an author to manually write their titles in lower case, but a major headache to work around these problems. This PR removes all instances of `\MakeLowercase` from the `mtheme`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,13 +61,26 @@ With option `nosectionslide`, no dedicated slide is produced when a new section
 starts. By default when using the `\section` command, a slide is created with
 just the title on it.
 
-The `nosmallcapitals` option supresses the usage of small capitals on
-title page, frame titles and dedicated section slides.
-
 Option `usetotalslideindicator` creates slide numbering in lower right corner
 in following format: #current/#total. By default, just current page number is
 printed.
 
+#### Title formatting
+
+The main title, section titles, and frame titles are all formatted according
+to the custom command `\mthemetitleformat`. By default, this is equivalent to
+`scshape` and sets the titles in small capitals, but you can change it in your
+preamble. For example:
+
+```latex
+\renewcommand{\mthemetitleformat}{}                       % no small capitals
+\renewcommand{\mthemetitleformat}{\scshape\MakeLowercase} % all small capitals
+\renewcommand{\mthemetitleformat}{\MakeUppercase}         % all capitals
+```
+
+Note that `\MakeLowercase` and `\MakeUppercase` can have unexpected behaviour
+in math mode, are disabled when `protectframetitle` is used, and cause crashes
+when an unprotected frametitle appears on a slide with `allowframebreaks`.
 
 #### Commands
 
--- a/beamerthemem.sty
+++ b/beamerthemem.sty
@@ -13,13 +13,11 @@
 
 \newif\if@useTitleProgressBar
 \newif\if@protectFrameTitle
-\newif\if@noSmallCapitals
 \newif\if@noSectionSlide
 \newif\if@useTotalSlideIndicator
 
 \@useTitleProgressBarfalse
 \@protectFrameTitlefalse
-\@noSmallCapitalsfalse
 \@noSectionSlidefalse
 \@useTotalSlideIndicatorfalse
 
@@ -37,7 +35,6 @@
   \PackageWarning{beamerthemem}{Unknown option `\CurrentOption'}%
 }
 
-\DeclareOptionBeamer{nosmallcapitals}{\@noSmallCapitalstrue}
 \DeclareOptionBeamer{nosectionslide}{\@noSectionSlidetrue}
 \DeclareOptionBeamer{usetotalslideindicator}{\@useTotalSlideIndicatortrue}
 
@@ -96,11 +93,7 @@
       \linespread{1.0}%
       \usebeamerfont{title}%
       \usebeamercolor[fg]{title}%
-      \if@noSmallCapitals%
-        \inserttitle%
-      \else%
-        \scshape\mthemetitleformat{\inserttitle}%
-      \fi%
+      \mthemetitleformat{\inserttitle}%
       \vspace*{0.5em}
     }}
     \fi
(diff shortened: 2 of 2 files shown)


matze:
> it's easy enough for an author to manually write their titles in lower case [...]

True, but both the TeX source as well as the TOC will look whacky. As a middle ground, maybe it's possible to make lower-casing optional?

benjamin-weiss:
As there are obviously many different opinions on this matter, maybe the best way is to provide 3 different alternatives as an option:

* Lowercase small-caps
* Uppercase regular
* no case adjustment

rchurchley:
Good points! I'd prefer not to add many new package options, since there's already quite a few of them and adding several more would overcomplicate the code. But there's a solution that I think will provide users the customization we want. This PR now replaces `\MakeLowercase` with a new command `\mthemetitleformat` which can be redefined by users to ensure the behaviour they want.

If this pull request is accepted, the default would be small caps with no case adjustment. This default minimizes unexpected behaviour in math mode and fixes the #20 crash. If they know it's safe, users can automatically lowercase their frame titles by adding 

```latex
\renewcommand{\mthemetitleformat}{\MakeLowercase}
```

to their preamble. Similarly, users with reason to prefer faked small caps can invoke 

```latex
\renewcommand{\mthemetitleformat}{\small\MakeUppercase}
```

Would this be a satisfactory compromise?

benjamin-weiss:
I like the idea. But if we use this solution I think we should also ged rid of the `noSmallCapitals` option and instead only use the hook.

rchurchley:
Sounds reasonable and now updated. In an ideal world we could just use `\setbeamerfont` the frametitle for most of these, but as far as I know you can't use that to set case.

matze:
Looks good and is an acceptable solution. Concerning the amount of options: that is true but IMHO they are less likely to cause problems for newcomers than redefining commands. But with the hook we can actually have best of both worlds: the options would simply set the hook to the appropriate value.

I will review the changes of this PR this afternoon and merge then ASAP.

benjamin-weiss:
You are right. Options are definitely easier for LaTeX beginners and newcomers to the theme. But I think with a rational explanation in the readme file this should not be a problem.

matze:
```latex
\renewcommand{\mthemetitleformat}{\MakeLowercase}
```
is not enough to get the current behaviour, for whatever reason one has to prepend `\scshape`. Anyway, @rchurchley could you update the README?

rchurchley:
Requiring a prepended `\scshape` was changed in d8500a1 after the initial PR — at @benjamin-weiss's suggestion, I replaced the `nosmallcapitals` option with the ability to `\renewcommand{\mthemetitleformat}{}`. I have now updated the README accordingly.

After implementing that change, I'm now on the fence as to whether it's the right move. It may be a "more beamer" solution to have `\mthemetitleformat` only change the case of titles and to reimplement the `nosmallcapitals` option in `beamerfontthememetropolis` using `\setbeamerfont`.

matze:
> I'm now on the fence as to whether it's the right move.

Since this theme is not yet in any way stable because of lacking proper versioning, I think we can change this if it becomes necessary. So for now, I'll merge and we will see.

jwhendy:
What's the definitive status on this? I'm wrestling with whether my failure to get lowercase titles is due to something with the documentation possibly being incorrect, a change like the above that might require an additional step, or simply because I haven't used `xelatex` before and I just don't have something setup correctly...

If I modify the `demo.tex` file like so, I still get uppercase frametitles:

```
\usetheme[frametitleformat=lowercase]{m}
```

I also tried using what I think was the outcome of the above, like this:

```
\usetheme[numbering=none]{m}
\renewcommand{\mthemetitleformat}{\MakeLowerCase}
```

That gives this error when I run `xelatex demo` from the `mtheme` folder:

```
! LaTeX Error: \mthemetitleformat undefined.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.4 \renewcommand{\mthemetitleformat}
                                     {\MakeLowerCase}
```

When I compile with regular `pdflatex`, I get lowercase titles but not the Fira font. When I compile with `xelatex`, I get the right font, but not the right casing. Thanks for any tips/suggestions!

If it's preferable that I setup my own new issue, I'm happy to do that. This seemed like the most relevant potential change to what the documentation instructs, so I thought others might look here and benefit from the answer vs. starting a brand new issue.

rchurchley:
I think we should probably rename some of the package options of this theme — the `lowercase` option could be more clear. By default, the metropolis theme sets lowercase letters in frame titles using small caps, Lɪᴋᴇ Tʜɪs. The `lowercase` option somewhat misleadingly forces the entire frame title to be small caps. If you want lowercase letters to actually be lowercase, just add the following to the preamble after you call the theme:

```latex
\setbeamerfont{frametitle}{shape=\upshape}
```

jwhendy:
Ah. Now I get it. Works like a charm, and I did totally misunderstand that option. Thanks for the quick assistance!

PR #59, opened 2015-05-28
Addresses issue with TeX capacity exceeded

### LINKED ISSUE #20: "TeX capacity exceeded" when using allowframebreaks

elliottd:
\begin{frame}[allowframebreaks] needs the title to be protected when
autobreaking because that is fragile.

There may be some [side-effects](http://texblog.net/help/latex/fragile.html)
when applying \protect to the title if used in combination with length and
counter commands.

Fixes #20

### DIFF
--- a/beamerthemem.sty
+++ b/beamerthemem.sty
@@ -278,10 +278,19 @@
 \nointerlineskip
 \begin{beamercolorbox}[wd=\paperwidth,leftskip=0.3cm,rightskip=0.3cm,ht=2.5ex,dp=1.5ex]{frametitle}
 \usebeamerfont{frametitle}%
-\if@noSmallCapitals%
-  \protect\insertframetitle%
-\else%
-  \textsc{\MakeLowercase{\protect\insertframetitle}}%
+\if@protectFrameTitle%
+  \protect%
+  \if@noSmallCapitals%
+    \insertframetitle%
+  \else%
+   \textsc{\MakeLowercase{\insertframetitle}}%
+  \fi%
+ \else%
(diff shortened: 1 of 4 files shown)


rchurchley:
This pull request fixes #20, but protecting `\insertframetitle` also prevents it from being affected by `\MakeLowercase`. I think this is probably inevitable — the crash seems to result from the interaction between `allowframebreaks` and `\MakeLowercase`. It may be less confusing to instead remove `\MakeLowercase` or ask the user to invoke it themselves. This solution is suggested as pull request #58.

matze:
Does this mean this PR is then obsolete?

elliottd:
I'll defer to @rchurchley about the true underlying cause of the problem here.

However, you may want to pull in the references frames in `demo.tex` and the references in `demo.bib` that demonstrate that allowframebreaks actually works.

matze:
Do you want to create a new commit for proper attribution?

rchurchley:
Looks like this can be merged now and #20 closed?

matze:
It must be rebased against master.

PR #60, opened 2015-05-31
Added 'noslidenumbers' option

ChipmunkMath:
I needed the ability to suppress slide numbering throughout my presentations, so I added a `noslidenumbers` option that can be invoked with the theme to omit page numbers in the footer.

I should mention that this is my first time working with a .sty file, so I might not have done this in the most elegant way possible. I had difficulty finding any documentation on how the `\if` command was working, so I couldn't figure out a way to add a third conditional. As such, I nested two `\if` commands to produce the desired result. If there's a better way to do this, that would be great to see, but I think being able to suppress page numbers is an important option for such a minimalist theme.

Thanks for the awesome theme! Looks great and excited to use it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,8 +69,6 @@ printed.
 
 Option `noslidenumbers` omits slide numbers entirely.
 
-<<<<<<< HEAD
-<<<<<<< HEAD
 #### Title formatting
 
 The main title, section titles, and frame titles are all formatted according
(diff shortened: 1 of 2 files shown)


matze:
Can you rebase your changes  (i.e. don't just merge the current master into yours)? The change itself looks good and useful to me.

ChipmunkMath:
Oof. This is pretty much my first time using git or GitHub, so there was a lot of fumbling around. Anyway, I think everything should be ready for merge.

matze:
Now, you ended up with a ton of commits :smile:

Anyway, I squashed and rebased cleanly for you and applied the changes directly to master, so I will close this PR.

I have to agree that Git's commands and repository model can be confusing but don't hesitate to ask if for such things. Also, it is a lot easier if you create a branch off of master before opening a pull request. That avoids some of the hassles you encountered.

PR #63, opened 2015-06-06
Reorganize and document package options

rchurchley:
This PR cleans up the package option declarations for improved readability. Despite the complicated diff, it only rearranges lines and adds comments; it does not add, change, or remove any code or output.


matze:
Looks a lot better now, thanks.

Issue #53, opened 2015-05-19
Support for subsections?


eddelbuettel:
I am converting/updating some existing presentations of mine.  Longer ones have subsections.

Any views on whether 
+ having subsections is desirable
+ subsections should show the section they are in
+ subsections should show the progressbar?

My latex-fu is a little weak; I tried to hack a quick macro together but it didn't quite pan out.  Thoughts?

matze:
*Disclaimer: this is purely subjective and my own opinion*

I never gave a talk longer than an hour and at least from my perspective just sections are usually enough to break up longer presentations. I know there is a tendency to break up a presentation into smaller topics if a theme encourage to do that (e.g. showing little dots and upcoming titles) but in my opinion this is just distracting and makes a presentation look very rigid.

Anyhow, I am not the one to decide how people structure their talks and in fact having smaller sections makes up for a nicer PDF table of contents. So here some thoughts:

> having subsections is desirable

As I said, people will agree or disagree :smile:

> subsections should show the section they are in

I would say so, otherwise it will probably just look like they are normal sections.

> subsections should show the progressbar?

I tend to say yes.

But it all boils down if you want to have subsections be handled like regular sections (i.e. with an inserted slide) or more subtle, for example by modifying the frametitle.

eddelbuettel:
> I never gave a talk longer than an hour.

Lucky you :)   I am about to give a two-day workshop in a few weeks; I somewhat regularly do three-hour tutorials.   Subsections are nice to have.  I also work on a few shorter talks and for those _m_ is already perfect.

As for the longer format and use of `subsection`: Currently I get one typeset, but it does not "mesh" with the rest of the _m_ theme.  The font is one size smaller and there is no title bar.  So I'd like to do better.  Currently I can either comment out the subsection use (via `\AtBeginSection[]{}`) or live with the not-so-pretty one.  Thoughts?

bgat:
I frequently give two-, three-, and four-day workshops.  Rather than use
subsections, I break my presentations up into individual files so that I'm
reminded to take breaks. It also lets me vary the order depending on
feedback during the event.

In my mind, subsections are evil...to the attendees.  :-)

b.g.


On Wed, May 20, 2015 at 7:11 AM, Dirk Eddelbuettel <notifications@github.com
> wrote:

> I never gave a talk longer than an hour.
>
> Lucky you :) I am about to give a two-day workshop in a few weeks; I
[...]

PR #66, opened 2015-06-10
Simplify colour theme for easy customization

rchurchley:
This PR implements an alternative to #64 to simplify `beamercolorthememetropolis.sty` and make it more easily customizable. Users would be able to change the colour scheme by setting any subset of four basic beamer colours:

```latex
\setbeamercolor{normal text}{...}
\setbeamercolor{alerted text}{...}
\setbeamercolor{structure}{...}
\setbeamercolor{block title}{...}
```

This approach incorporates the semantic improvements suggested by #64 while keeping descriptive names for the primitive colours defined by `\definecolor`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,11 +40,11 @@ presentations, you can run the following command
 
 #### Package options
 
-To use any of options below, call them when invoking *mtheme* in the preabmle of
+To use any of options below, call them when invoking *mtheme* in the preamble of
 the slides, i.e.
 
 ```latex
-  \usetheme[<options>]{m}
+\usetheme[<options>]{m}
 ```
 
 * The `usetitleprogressbar` option adds a thin progress bar similar to the
@@ -66,6 +66,25 @@ the slides, i.e.
   in following format: #current/#total. By default, just current page number is
   printed.
 * Option `noslidenumbers` omits slide numbers entirely.
+
+#### Color customization
+
+The included `metropolis` color theme is used by default, but its colours can
+be easily changed to suit your tastes. All of the theme's styles are defined
+in terms of three beamer colors:
+
+- `normal text`     (dark fg, light bg)
+- `alerted text`    (colored fg, should be visible against dark or light)
+- `example text`    (colored fg, should be visible against dark or light)
+
+An easy way to customize the theme is to redefine these colors using
+
+```latex
+\setbeamercolor{ ... }{ fg= ... , bg= ... }
+```
+
+in your preamble. For greater customization, you can redefine any of the other
+colors in `beamercolorthememetropolis`, including `progress bar`.
 
 
 #### Title formatting
--- a/beamercolorthememetropolis.sty
+++ b/beamercolorthememetropolis.sty
@@ -106,8 +106,11 @@
 %
 
 \setbeamercolor{frametitle}{use=palette primary, parent=palette primary}
-\setbeamercolor{progress bar}{use=alerted text, parent=alerted text}
-\setbeamercolor{progress bar incomplete}{use=structure, parent=structure}
+\setbeamercolor{progress bar}{%
+  use={alerted text, structure},
+  fg=alerted text.fg,
+  bg=structure.fg
+}
 
 \setbeamercolor{background canvas}{use=normal text, parent=normal text}
 \setbeamercolor{titlelike}{use=normal text, parent=normal text}
--- a/beamercolorthememetropolis.sty
+++ b/beamercolorthememetropolis.sty
@@ -3,19 +3,23 @@
 % Copyright 2014 Matthias Vogelgesang
 % Licensed under CC-BY-SA 4.0 International.
 %
-% The initial template comes from the HSRM beamer theme by Benjamin Weiss, which
-% you can find at https://github.com/hsrmbeamertheme/hsrmbeamertheme.
-%
 
 \ProvidesPackage{beamercolorthememetropolis}
 
 
-%}}}
-%{{{ --- Options ----------------------
+% Options
+% =======
+
+% Option: blockbg
+%         applies a gray background to blocks
 
 \newif\if@beamer@metropolis@blockbg
 \@beamer@metropolis@blockbgfalse
-\DeclareOptionBeamer{blockbg}{\@beamer@metropolis@blockbgtrue}
+\DeclareOptionBeamer{blockbg}{
+    \@beamer@metropolis@blockbgtrue
+}
+
+% Option: ...
 
 \DeclareOptionBeamer*{%
   \PackageWarning{beamercolorthememetropolis}{Unknown option `\CurrentOption'}%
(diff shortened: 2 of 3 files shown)


matze:
For some reason bullet points get lost. Other than that, it looks like a good base to start developing new color themes :+1:

benjamin-weiss:
> For some reason bullet points get lost.

Not only the bullet points. Adding the following should fix most of the errors this PR introduces.

    \setbeamercolor{section in toc}{use=normal text, fg=normal text.fg}
    \setbeamercolor{item}{use=normal text, fg=normal text.fg}
    \setbeamercolor{caption name}{use=normal text, fg=normal text.fg}
    \setbeamercolor{qed symbol}{use=normal text, fg=normal text.fg}

Also the block body for the example block and the alert block are not defined yet.

    \setbeamercolor{block body alerted}{parent=block body}
    \setbeamercolor{block body example}{parent=block body}

As mentioned earlier https://github.com/matze/mtheme/pull/64#issuecomment-110467158 I think we really should also define the example text color as one of the green shades that are already in the original [Paletton](http://paletton.com/#uid=7050t0kkJkJsntwoyp6gYgoddc4), e.g.:

    \definecolor{mLightGreen}{HTML}{14B03D}
    \setbeamercolor{example text}{mLightGreen}

And once again https://github.com/matze/mtheme/pull/64#issuecomment-110395859: I do not think it is really useful nor advisable to use `\ProvidesPackage{•}` in a color theme. After a quick test where I removed it and loaded the color theme twice I had no errors. So even the package option shouldn't be a problem.

benjamin-weiss:
And another minor problem. Although it seems to be a good idea at first, defining `block body` outside of the if condition is a problem. Because `bg=block title.bg!50` leads to a slightly lighter `block body` than the title (`#FCFCFC` instead of `#FAFAFA`). So I would propose to implement it like that.

    \if@beamer@metropolis@blockbg
      \setbeamercolor{block title}{%
        use=normal text,
        fg=normal text.fg,
        bg=normal text.bg!80!fg
      }
      \setbeamercolor{block body}{
        use=block title,
        bg=block title.bg!50
      }
    \else
      \setbeamercolor{block title}{
        use=normal text,
        parent=normal text
      }
      \setbeamercolor{block body}{
        use=normal text,
        parent=normal text
      }
    \fi

rchurchley:
> For some reason bullet points get lost.
> Not only the bullet points.

I misunderstood what beamer uses `structure` for; I should have been using something like `palette primary` instead. This is now fixed.

> Also the block body for the example block and the alert block are not defined yet... I think we really should also define the example text color as one of the green shades that are already in the original Paletton.

Good catch and great idea. Both suggestions are now incorporated.

> Because bg=block title.bg!50 leads to a slightly lighter block body than the title (#FCFCFC instead of #FAFAFA).

I woke up this morning realizing this was a problem. I think the bigger problem is using `bg=block title.bg!50` in the first place — the colour mix needs to be defined relative to `normal text.bg`. This fixes the problem you point out as well as another: if the user defines a dark theme by redefining the background of `normal text`, we expect `block body` to be *darker* than `block title`, not lighter.

benjamin-weiss:
Good work so far! Lets go to the next proposal: At the moment it is not possible to change the color of the progress bar independently from the alert color without editing the `beamerthemem.sty`. So I think we should define a separate color for the progress bar which is – for the metropolis theme – dependent on the alert color.

rchurchley:
> And once again #64 (comment): I do not think it is really useful nor advisable to use \ProvidesPackage{•} in a color theme.

I'm not sure I understand this proposal. (All the beamer colour themes I know of use some variant of `\ProvidesPackage`.) Is it because you want to `\input` the colour theme mid-presentation or something?

rchurchley:
> So I think we should define a separate color for the progress bar which is – for the metropolis theme – dependent on the alert color.

Also a great proposal and now implemented.

benjamin-weiss:
I would have combined it into one color with foreground and background.

    \setbeamercolor{progressbar}{%
        use={alerted text, structure},
        fg=alerted text.fg,
        bg=structure.fg
    }

But your way is also fine.

rchurchley:
@benjamin-weiss Your way is way better, now updated.

benjamin-weiss:
> I'm not sure I understand this proposal. (All the beamer colour themes I know of use some variant of \ProvidesPackage.) Is it because you want to \input the colour theme mid-presentation or something?

Just looked again at some of the [build in color themes](https://bitbucket.org/rivanvx/beamer/src/3dd33d75f1b859b0918b4aed8d3f1451fddbdf99/base/themes/color/?at=default) could not find one with `\ProvidesPackage`. If I understand `\ProvidesPackage` right, its purpose is to warn you when you include the file twice. But why should it do that in our case? There is no harm in loading in twice. And a possible use case is the one I already explained [here](https://github.com/matze/mtheme/pull/64#issuecomment-110395859).

    \usetheme{m} % loads metropolis color theme
    ...
    \section{Section with stock colors}
    ...
    \usecolortheme{gotham}
    \section{Section with different colors}
    ...
    \usecolortheme{metropolis}
    \section{Section with stock colors}

benjamin-weiss:
Another use case which comes to mind is to enable/disable `blockbg` mid presentation.

rchurchley:
Interesting. The [themes in the latest released version of beamer (3.36)](https://www.ctan.org/tex-archive/macros/latex/contrib/beamer/base/themes/color) all use a `\ProvidesPackageRCS` command, which invokes `\ProvidesPackage` with some version control stuff automatically added. But as your link shows, they were removed from the development version in March. That was news to me!

There are several other purposes to `\ProvidesPackage` — declaring versions for a package, for example — but we may not be using any of them. Perhaps it is safe to remove it after all. But I'm not sure if we actually get the benefits you describe (it would be great if it did!) as your example code doesn't seem to work for me.

Removing `\ProvidesPackage` allows us to call the same theme twice within the preamble, but `\usecolortheme` doesn't work after `\begin{document}` has been called. Your example code throws 

```
LaTeX Error: Can be used only in preamble. [\usecolortheme{gotham}]
```
even when `beamercolorthemegotham.sty` is an empty theme with no `\ProvidesPackage`.

benjamin-weiss:
With the changes `item`, `section in toc`, `caption name` and `qed symbol` are now in the `structure` color, hence `mDarkTeal!50`. Do we want that? I'm inclined to use the original `mDarkTeal` in those cases. Especially for a list in a `blockbg` block the contrast is too low imho.

item_screenshot

So I think we should incorporate the changes I proposed earlier despite the better usage of `structure` you implemented.

    \setbeamercolor{section in toc}{use=normal text, fg=normal text.fg}
    \setbeamercolor{item}{use=normal text, fg=normal text.fg}
    \setbeamercolor{caption name}{use=normal text, fg=normal text.fg}
    \setbeamercolor{qed symbol}{use=normal text, fg=normal text.fg}

benjamin-weiss:
> Removing `\ProvidesPackage` allows us to call the same theme twice within the preamble, but `\usecolortheme` doesn't work after `\begin{document}` has been called.

Ok, I just assumed that would be possible. But if not, we can obviously keep `\ProvidePackage`.

benjamin-weiss:
Also the bibliography entry colors have too little contrast. Compare the code of the `beamercolorthemedefault.sty`:

    \setbeamercolor{bibliography entry author}{use=structure,fg=structure.fg}
    \setbeamercolor{bibliography entry title}{use=normal text,fg=normal text.fg}
    \setbeamercolor{bibliography entry location}{use=structure,fg=structure.fg!65!bg}
    \setbeamercolor{bibliography entry note}{use=structure,fg=structure.fg!65!bg}

bib_screenshot

I would propose to change also all the bibliography entries to `{use=normal text, fg=normal text.fg}` and implement different font styles for the `author`, `title` and so on

benjamin-weiss:
What do you think about that?

bib_screenshot

**`beamerfontthememotropolis.sty`**

    \setbeamerfont{bibliography entry author}{family=\Light, size=\normalsize}
    \setbeamerfont{bibliography entry title}{family=\Medium, size=\normalsize}
    \setbeamerfont{bibliography entry location}{family=\Light, size=\normalsize}
    \setbeamerfont{bibliography entry note}{family=\Light, size=\small}

**`beamercolorthememotropolis.sty`**

    \setbeamercolor{bibliography entry author}{use=normal text,fg=normal text.fg}
    \setbeamercolor{bibliography entry title}{use=normal text,fg=normal text.fg}
    \setbeamercolor{bibliography entry location}{use=normal text,fg=normal text.fg!80!bg}
    \setbeamercolor{bibliography entry note}{use=normal text,fg=normal text.fg!50!bg}

rchurchley:
At this point, it seems best to just declare `structure` to be the same as `normal text`, so I've made that change. How does it look now?

benjamin-weiss:
That was probably a good idea. Keeps the code clean. With your changes I think we can dismiss the adjustments I proposed for the `beamercolorthememetropolis.sty` file, but I would still use the changes I made in the `beamerfontthememetropolis.sty` file.

bib_screenshot_1

bib_screenshot_2

rchurchley:
Since this PR currently only addresses colour issues, perhaps we could tackle the fonts in a different issue/pull request?

benjamin-weiss:
Yes, makes sense.

Just went over the `beamercolorthemedefault`file again and the only thing we haven't looked at yet, as far as I see it, are the beamer buttons. But I think they look fine already (at least color-wise, the triangle looks a bit off).

button_screenshot

benjamin-weiss:
The line on the title page is still dependent on the `alert text` color. We should either rename the `progressbar` color to something like `highlights` and use this color on the progressbar *and* the line on the title page or define a new color for that.

And the comment *Base colors* should be copied to the README file.

rchurchley:
Documentation added to the README, and colours shuffled around a bit to reflect their dependency structure. I added a new `title separator` colour with `progress bar` as a parent, so users can redefine `progress bar` and get a matching title separator for free, but also have the option of changing them separately.

benjamin-weiss:
Sounds good. I also tested everything with an "inverse" theme (dark background and light foreground). There also everything seems to be working as expected.

[demo_inverse.pdf](https://github.com/benjamin-weiss/mtheme/raw/new-themes/demo_inverse.pdf)

benjamin-weiss:
Speaking of "inverse" themes. What do you think about an `invertcolors` option. Implemented like this:

    \if@beamer@metropolis@invertcolors
      \setbeamercolor{normal text}{%
        fg=black!2,
        bg=mDarkTeal
      }
    \else
      \setbeamercolor{normal text}{%
        fg=mDarkTeal,
        bg=black!2
      }
    \fi

rchurchley:
Sounds like a reasonable idea. Perhaps you'd like to take credit for it with a pull request of your own? I probably shouldn't let this branch get *too* far away from `master` before @matze gets a chance to merge it :)

benjamin-weiss:
Alright, it is definitely a good point to merge right now.

The proposed option could anyway raise a discussion. 	If I remember rightly, we already had a discussion about too many options. And it can be done very easily without the option.

matze:
The changes look good to me, thanks for the awesome work to both of you. But before it gets lost when closing the PR, I want to bring up [this](https://github.com/rchurchley/mtheme/commit/db8014c00bf9932362d3ec790864af3f0963b06f) comment into discussion. In essence, I'd like to have separate progress bar colors for frametitle and section title or, if that is too much, move away from the orange.

benjamin-weiss:
> Would it be possible to have two independent progress bar colors for frame title and sections?

It shouldn't be a problem at all. I would propose a definition like that:

    \setbeamercolor{progress bar}{%
      use=alerted text,
      fg=alerted text.fg,
      bg=normal text.bg!50!fg
    }
    \setbeamercolor{progress bar section}{%
      use=progress bar,
      parent=progress bar,
    }
    \setbeamercolor{progress bar frame}{%
      use=progress bar,
      parent=progress bar,
    }

With a system like that you can still change them independently, but if you do not care and want to have them changed together it is as easy as before.

rchurchley:
> It shouldn't be a problem at all.

It will take some careful implementation in `beamerthemem.sty`. Sadly, beamer isn't smart enough to automatically detect when a progress bar is on a section page or normal frame. I can think of a few different options to overcome this, but each has its own tradeoffs and implications for how we extend the theme in the future (e.g. to resolve #53). I'll have to give this some thought. Perhaps this topic could be spun off into a separate Issue?

> or, if that is too much, move away from the orange.

I don't have any strong opinions about the particular values of the default colours: I override them anyways to optimize my presentations for lousy projectors :) As far as I'm concerned, you're the boss!

matze:
Alright, I will take care of this in a separate commit then. Thanks again!

PR #72, opened 2015-06-11
added invertcolors option

benjamin-weiss:
This PR implements a option to switch the foreground and background color as discussed in #66.

Maybe this could be useful for some people.


matze:
Could you add an entry to the README?

benjamin-weiss:
Ok, I will do that. I just first wanted to know if this is a feature that is really wanted.

matze:
Yes, definitely. Especially with dark images (e.g. scientific imaging) the contrast is much better. However, I noticed one issue. Right now it's basically only supporting the default color theme, so we should document or otherwise state that this won't work with custom color themes.

benjamin-weiss:
> Right now it's basically only supporting the default color theme

I thought about that too. A global solution would be much nicer. But sadly 

    \setbeamercolor{normal text}{%
      use=normal text,
      fg=normal text.bg,
      bg=normal text.fg
    }

does not work. And other solutions to make it a global I can think of will make the code way to messy.

rchurchley:
I don't think an `inverted` option has an unambiguous expected behaviour for custom colour themes anyways. If I `\setbeamercolor{normal text}{fg=blue}` with the `inverted` option, am I trying to invert a tweaked colour theme or tweak the inverted colour theme?

In theory, we could create an `\invertcolors` command which swaps the foreground and background of `normal text` via temporary variables. But I don't think that's a good idea either, since the user could decouple another Beamer colour from `normal text` and get confused when it is not inverted.

I think it's fine to keep this behaviour as it is. Perhaps the option's name could be changed to `dark` instead of `inverted` to make it more clear that it does not invert the user's colour changes.

benjamin-weiss:
>  Perhaps the option's name could be changed to `dark` instead of `inverted` to make it more clear that it does not invert the user's colour changes.

That is a good point. What do you think of the name `darkcolors`? That is even more descriptive.

rchurchley:
`darkcolors` is more descriptive, while `dark` is probably easy to guess/remember. I don't have a preference either way.

benjamin-weiss:
@matze Did you just forget to merge or is there something still not right with this PR?

matze:
Sorry, it just got lost. You are too productive and overwhelm me with PRs :smile:

matze:
By the way, you should maybe rebase the next time, all those "Merge pull request" commits are not really nice too look at.

benjamin-weiss:
>  ... you should maybe rebase the next time ... 

Ok, I promise to improve. But a quick question (as I'm still quite new to git) regarding this matter. What should I do if I have multiple open PRs? I can rebase at the time I create the PR. But if you pull one of them we would still have the "Merge pull request" problem with the others, right? So to prevent it you would have to accept the first PR, then I would have to rebase, then you would have to accept the next PR ...

Or is there an easier solution?

matze:
The easiest would be to rebase against master right before we intend to merge (and maybe squash) and then force-push the changes. In the other branches you would then rebase against the updated master that then includes the clean merge.

PR #73, opened 2015-06-11
removed document class option compress

benjamin-weiss:
This PR removes the unused `compress` option. No big thing. But it annoys me every time I see it. ;)


matze:
Well, okay :smile:

Issue #69, opened 2015-06-11
Handling plot "themes"


matze:
Right now the pgfplots theme is hardcoded into the Beamer theme itself. This is not good for various reasons. First of all, the color definitions might clash with a different color theme (once #64 is merged this will be much more of an issue). And even worse: pgfplots might not be available.

Ideally we would have a command to load a pgfplot theme just like `\usecolortheme`, i.e. something like `\usepgfplottheme`. In the best case pgfplots would provide that, but I don't know about such thing.
This command could also take care to not issue `pgfplotset` in case pgfplots is not available.

rchurchley:
I've had my eye on this too :+1: 

Even if `\usepgfplottheme` doesn't exist, we can still bundle off the pgfplots code into a little sub-package and include it with `\RequirePackage`. We can then call the package only when `pgfplots` is available, much like we currently `\usefonttheme` only under XeLaTeX or LuaLaTeX. [This answer](http://tex.stackexchange.com/questions/15306/conditional-if-a-package-is-available) appears to show how to do that.

benjamin-weiss:
From #88:
> Right now the pgfplot theme is tied to a particular color. Ideally we would have a pgf color theme, which could actually come from the color theme itself in order to provide consistency, and a structural theme, i.e. all those axes, legend and grid customizations.

I think it would be sufficient to define the set of colors in the `plotcyclelist`. Axes, legend, grid and so on seems to be working just fine already.

![screenshot_dark_plot](https://cloud.githubusercontent.com/assets/9158719/8201852/0ad75ae8-14d5-11e5-9912-981df519b3d9.png)
  
Screenshot created using the following color adjustments:

    \definecolor{mBloodRed}{HTML}{FF0000}
    \definecolor{mYellowBastard}{HTML}{EAF02C}

    \setbeamercolor{normal text}{%
      fg=white,
      bg=black!90
    }

    \setbeamercolor{alerted text}{%
      fg=mBloodRed
    }

    \setbeamercolor{example text}{%
      fg=mYellowBastard
    }

rchurchley:
> Axes, legend, grid and so on seems to be working just fine already.

Which version of the theme are you working under? I briefly broke the behaviour of `mbarplot` on my branch, so if you have that checked out you may be seeing the default `beamer` colour-choosing behaviour. Perhaps this is a sign to drop the Tol colours from the theme altogether?

benjamin-weiss:
[...]

Issue #68, opened 2015-06-11
Distribution of the theme via CTAN


benjamin-weiss:
The manual installation of the theme is annoying and maybe difficult for beginners. For example I have to install the theme on every PC and for every user separately. So I think it would help a lot to distribute the theme as a package via [CTAN](http://ctan.org). It would also reach more users, as a lot of the TeX community is probably not active on GitHub. Also this would make other ways of distribution like the .deb package obsolete. We would have only one way of distribution left for everyone.

As far as I looked into it, this is what has to be done:

* Add dates to `\ProvidesPackage` commands
* Create a Makefile rule which creates the zip file to be uploaded to CTAN
* Create a documentation PDF
    - This is not mandatory, but very useful. If I want to get infos about a package I do not know I always use `texdoc {packagename}`.
    - The easiest way do do that is probably to adjust the README.md that pandoc creates a nicely formatted PDF.
* In this context it would be useful to rename the theme from `m` to something more self-explanatory like `modernbeamer`.
* Upload it to CTAN [here](http://ctan.org/upload/)

If I get it right the folder structure of the zip file should look like this in our case:

    mtheme.zip:
      mtheme/
        README   % to use as a changelog
        doc/
          mtheme.md   % renamed README.md
          mtheme.pdf  % using pandoc created documentation
        tex/
          beamercolorthememetropolis.sty
          beamerfontthememetropolis.sty
          beamerthemem.sty

Or for a very simple package without documentation just:

    mtheme.zip:
      mtheme/
        README.md
        beamercolorthememetropolis.sty
        beamerfontthememetropolis.sty
        beamerthemem.sty

Does anybody have experience with CTAN and can confirm my infos or can give some more insights?

ChipmunkMath:
I agree that distributing the theme through CTAN is a good idea: it would increase access to a beautiful theme. However, how easy is it to update the theme if it's on CTAN? Will it slow down the ability to push out changes? Right now, it still seems like there's regular modifications. If distributing via CTAN will slow that process down, this should probably be put on hold for a couple of months, at least until the theme has stabilized further.

matze:
> I agree that distributing the theme through CTAN is a good idea [...]

Indeed. I didn't upload to CTAN yet because I wasn't willing to spend much time on bringing everything up to their standards. But from what I can tell, it doesn't look horribly difficult.

[...]

PR #74, opened 2015-06-12
Documented LaTeX Format

benjamin-weiss:
As promised I looked into the *Documented LaTeX Format* and I think this is definitely the way to go. It is all a bit confusing at first, but you get the hang of it quite fast.

At the moment everything is still very basic. I also removed a lot of the old comments as they were making trouble and I wanted to focus on getting the structure right and get everything to work. And it is probably anyway reasonable to start with a clean slate and define a consistent system regarding how the comments should look like. All in all it is still a bit messy, but it works so far.

The `sty` files can be extracted via

    latex mtheme.ins

And the documentation can be created for example with

    latexmk -latex=xelatex mtheme.dtx

So, what do you think?


matze:
Looks like the way to go*. I guess when this progressed far enough everything in `package` will move to the top directory?

* Though I probably have to get used to the inline code.

benjamin-weiss:
> I guess when this progressed far enough everything in package will move to the top directory?

Haven't thought of it yet. But I think this makes sense.

benjamin-weiss:
There are some issues where I'd like to have some input:

1. As I already mentioned before in #68 I think when we release the theme on CTAN, the theme should be renamed. `m` is not very self-explanatory. Do you agree? If so, what would be a good name? `modern`?
2. Do we want to append the whole source code in the pdf documentation? It seems to be quite common for smaller packages. I tend to say yes we do that. This will affect the decision how we format our comments. [minted](https://github.com/gpoore/minted/) does it kind of in the way I have in mind. 
3. What information do we keep in the `README.md` file after the pdf documentation is finished? And shall it be static or automatically created out of the `mtheme.dtx`? I tend to say we make it static, keep only a short overview with screenshot, link to the package documentation and add the license information. Maybe – but not necessarily – also a change log. 
4. The codebase of the theme is growing. Do you think it makes sense to adapt the folder structure? Add a `color`, `font` (I'm thinking about providing different font themes, e.g. a serif one) and a `demo` or `examples` folder?
5. Which build rules do we keep in the `Makefile` after we finished converting the theme to the *Documented LaTeX Format*? And which do we need to add? Are the docker rules still useful when the theme is available on CTAN (I know nearly nothing about docker)? Why is there after all a separate `build.sh`?

matze:
> As I already mentioned before in #68 I think when we release the theme on CTAN, the theme should be renamed. m is not very self-explanatory. Do you agree? If so, what would be a good name? modern?

`modern` implies that all other themes are … not modern. But if we don't come up with a better name and this will help conforming to CTAN's rules, let it be it.

> Do we want to append the whole source code in the pdf documentation?

Personally, I am not a huge fan of these kind of docs because I easily lose focus of what's important. But you are right, the theme is small enough that we could cover the whole source.

> What information do we keep in the README.md file after the pdf documentation is finished? And shall it be static or automatically created out of the mtheme.dtx? I tend to say we make it static, keep only a short overview with screenshot, link to the package documentation and add the license information. Maybe – but not necessarily – also a change log. 

I agree with the overview, doc link and license but I'd keep the change log inside the real docs.

> The codebase of the theme is growing. Do you think it makes sense to adapt the folder structure?

Well, technically nothing speaks against it and in fact the official themes are split up like that anyway, so go for it :smile:

> Which build rules do we keep in the Makefile after we finished converting the theme to the Documented LaTeX Format?

At least a `build` target for the demo and an `install` target for local installation. Also the docs should be build with `make docs` or `make pdf`. Last but not least, `make clean` is pretty handy.

> Are the docker rules still useful when the theme is available on CTAN (I know nearly nothing about docker)?

For the nearest future, we should keep it. Having it at CTAN does not mean it will be available on all distributions immediately. I still have painful memories using TeXLive 2009 on Ubuntu 12.04 …

> Why is there after all a separate build.sh?

Good question, but it probably was the most convenient way for @awalterschulze to set it up.

rchurchley:
> what would be a good name?

If `modern` is too generic, `metropolis` could be a good choice. It's distinct, already has a history with this project, follows the "cities" naming scheme of beamer themes (if that carries any weight), and is much more google-able.

> Do we want to append the whole source code in the pdf documentation?

I don't have an informed opinion on this yet — give me a couple days to get familiar with the `.dtx` format.  I assume that most people who are interested enough to look at the source code will also be interested in making modifications. Therefore, I'd prefer whichever option makes the `.dtx` source most readable and maintainable; I don't want the editable source to suffer in the cause of adding implementation details to the PDF documentation.

> What information do we keep in the README.md file after the pdf documentation is finished? I tend to say we make it static, keep only a short overview with screenshot, link to the package documentation and add the license information.

I believe the README should be kept static and only contain a brief description. It might(?) be used by CTAN as the description of the package, so I'm not sure if screenshots will work.

> Do you think it makes sense to adapt the folder structure? 

A folder structure may or may not be useful, though I don't know how many files we intend to add or what the optimal structure would be.

rchurchley:
I'll have to learn more about Documented LaTeX over the weekend to provide more substantive comments, but I do have a couple quick questions/issues to raise over the current PR.

First, the line

```latex
The Current Maintainer of this work is Michael Vogelgesang.
```

is a little out-of-place legally speaking. The `mtheme` is currently released under a Creative Commons license, under which "Current Maintainer" doesn't have a specific meaning. But it does have a specific meaning under the LaTeX Project Public Licence, which is commonly used by CTAN packages. The presence of this line therefore could cause a tiny amount of confusion over which license applies or has historically applied.

Of course, we could release the `mtheme` under the LPPL if we wanted. (As as the primary copyright holder, @matze should probably be the one to make the final decision and the particular commit changing the license.) What do you guys think?

Secondly (though I haven't gotten a change to test this) it looks like from the code that the documentation is set in Fira as well. Does this mean people without Fira will not be able to compile the documentation?

benjamin-weiss:
> `modern` implies that all other themes are … not modern. But if we don't come up with a better name and this will help conforming to CTAN's rules, let it be it.

Well, actually the package name for CTAN and the beamer theme name are two different things. Right now I already renamed the package to `mtheme`, with what I'm not so pleased either. So we definitely could release it the way it is. But I think it would be nice if the package name is indeed the same as the beamer theme name.

> Personally, I am not a huge fan of these kind of docs because I easily lose focus of what's important.

The thing is that traditionally the created `sty` files are intentionally stripped of all comments and even blank lines. You can prevent it ("['^^A' can be used instead of '%' to introduce a comment](https://www.tug.org/TUGboat/tb29-2/tb92pakin.pdf)"), but that is not very convenient. So for most of the users the documentation is the place to look at a commented source. I'm not so convinced of this system either, but I'm a huge fan of standards and this feels to me like an unofficial standard.

> I agree with the overview, doc link and license but I'd keep the change log inside the real docs.

Does your comment also imply we should use a static `README.md`? 

> At least a `build` target for the demo and an `install` target for local installation. Also the docs should be build with `make docs` or `make pdf`. Last but not least, `make clean` is pretty handy.

Sounds reasonable. CTAN package creation could also be incorporated into the Makefile. Something like `make zip` or `make ctan`.

> For the nearest future, we should keep it.

Ok.

rchurchley:
And one last (potentially off-topic) question of git etiquette. I've got a few more changes I'd like to make to the existing code base (e.g. resolving #69). Should I finish and pull request those now or wait until the conversion to `.dtx` is finished?

benjamin-weiss:
> `metropolis` could be a good choice

I like that.

> I'd prefer whichever option makes the `.dtx` source most readable and maintainable; I don't want the editable source to suffer in the cause of adding implementation details to the PDF documentation.

As far as I can see it so far. This should not be a problem.

> It might(?) be used by CTAN as the description of the package, so I'm not sure if screenshots will work.

I would say we provide a different README file for the CTAN package. Compare again with the [minted](https://github.com/gpoore/minted) package.

> I don't know how many files we intend to add or what the optimal structure would be.

I'm also still unsure how the optimal structure would be.

> `The Current Maintainer of this work is Michael Vogelgesang.`

Ups did I gave @matze a new name. At least it is still a man's name. ;)

> The presence of this line therefore could cause a tiny amount of confusion over which license applies or has historically applied.

Good points. I'm not really familiar with the LPPL. But this could really a good idea to change the license. But as you said the final decision should lie with @matze.

> Does this mean people without Fira will not be able to compile the documentation?

Right now yes. But we can always implement a similar system we already have in the theme.

benjamin-weiss:
> And one last (potentially off-topic) question of git etiquette. I've got a few more changes I'd like to make to the existing code base (e.g. resolving #69). Should I finish and pull request those now or wait until the conversion to `.dtx` is finished?

That is also a good point. All in all my current implementation already works. So maybe it is a good idea to merge at this point. Keep the old code base for reference and start to implement new changes on the `dtx` file only. And when the new implementation is finished we just remove the old code base and move the new implementation to the top directory as proposed by @matze?

This would also make it easier to collaborate on the new `dtx` implementation and not all changes have to go through me and my PR here.

rchurchley:
Is is possible for @matze to accept @benjamin-weiss's PR as a new branch on the project and only merge with `master` when it's ready?

benjamin-weiss:
That feels more like a git way to do things. Then we can also immediately start developing on the top directory.
It seems like it is not possible directly (https://github.com/isaacs/github/issues/18), but we can alway close this PR and just create a new one.

matze:
I will just comment in general now:

1. Yes, I am totally fine with switching from CC-BY-SA to LPPL. CC-BY-SA was the closest that was neither too liberal nor to restrictive, but on the other hand it is in any case more suitable for artwork and assets than for a theme.
2. Yes, IMHO the README should be static.
3. So what exactly do you prefer now? 1) Merge this. 2) Push this as an "official" branch (though I doubt you are then able to push). 3) Wait and merge commits based on the non-dtx'd theme?

benjamin-weiss:
> So what exactly do you prefer now?

Every option has its pros and cons. I tend to an official development branch. But yes you are right. Then you would have to manage all the PRs, I wouldn't be able to simply push my changes.

matze:
Okay, before we make this overly complicated: just finished the bare minimum and I will merge, so that the rest can be based on top of your changes. In the end, one has to do a little work either you by integrating the other commits or the others by merging it to the dtx version.

benjamin-weiss:
Ok, so we do it like I proposed at first. Keep the package folder, merge, only allow new changes on the `dtx` files to prevent unnecessary merges and move it later to the top directory? That's what you meant?

rchurchley:
If it works, then personally I think you should just `.dtx`-ify the existing files and we can rebase our changes against that. We're going to have to do that anyways, so we might as well do that now rather than confuse everybody by having two copies of everything in the same project.

matze:
> That's what you meant?

That what @rchurchley is much better expressing at ;-)

benjamin-weiss:
Still a bit unsure if that's what you meant. But I hope I did it like you wanted.

matze:
Yes, that's good. Thanks for the work!

PR #75, opened 2015-06-13
Added new font theme for thicker, easier to read letters.

ChipmunkMath:
Put together a new Beamer font theme: `dailyplanet`. Use with `\usefonttheme{dailyplanet}` sometime after `\usetheme[<options>]{m}` in the document header. This changes the normal font from Fira Sans `Light` to primarily using Fira Sans `Book`. Similar changes were made to the rest of the supporting font choices.

fontprovingground

This font theme was made with high readability in mind. While I prefer the aesthetic quality of the default `metropolis` font theme, my use case necessitates readability on phone screens. The new theme also helps guarantee readability in other challenging situations, like large auditoriums and/or small projectors.

Some minor issues to deal with before this is ready to be merged:
- **Bibliography Fonts**: I really struggled with getting a bibliography to work in Beamer using xelatex and ultimately gave up. I tried to structure the bibliography entries around the same pattern that the default theme has, but someone else will have to verify that it isn't terrible.
- **Theme Name**: Is `dailyplanet` an acceptable name? It's a fun call back to the theme's name, but it isn't very descriptive. Then again, that issue might be mitigated in the future if we have documentation showing illustrations of each theme.
- **ReadMe**: Haven't touched the ReadMe yet. Wasn't sure if it was best to create another subsection (like the current *Color customization*), or list it somewhere else. If we create a subsection on *Font Customization*, should we also give some general pointers to the larger idea of using any font you want by making a custom font theme? Or just say how to use `dailyplanet` and leave it at that?

Finally (and I'm not sure if this is the proper GitHub etiquette or I should open an Issue for these things), the research I wound up doing prior to creating this new font theme led me to the discovery of these two packages. I thought they might be of interest if you didn't already know of them:
- [fira](http://www.ctan.org/pkg/fira) // **Fira Sans Fonts**: Provides Fira Sans and Fira Mono fonts, and works outside of xelatex/lualatex. I was playing around with it and typesetting just in plain pdflatex. Seems to be in pretty good shape, although it's important to make sure you're updated to the most recent version (2015-05-23) which fixes some major gaps.
- [newtxsf](http://ctan.org/pkg/newtxsf) // **Sans Math Fonts**: Provides sans-serif math fonts, such as Greek. Right now, I think there's a bit of an issue in `metropolis` with using Greek symbols, since they feel so different from Fira Sans. While this is less of an issue when using `Fira Sans Light`, it becomes especially noticeable if you're using the `dailyplanet` font theme (Take a look at \sin(\theta) in the image above). Using the `newtxsf` package gives a much more pleasing match to my eye.

Both of the above packages seem to be quite new (released after `mtheme` was posted), so I wanted to make sure I pointed them out.

### DIFF
--- a/beamerfontthemedailyplanet.sty
+++ b/beamerfontthemedailyplanet.sty
@@ -1,0 +1,55 @@
+% Beamer mtheme
+%
+% Copyright 2014 Matthias Vogelgesang
+% Licensed under CC-BY-SA 4.0 International.
+%
+% The initial template comes from the HSRM beamer theme by Benjamin Weiss, which
+% you can find at https://github.com/hsrmbeamertheme/hsrmbeamertheme.
+%
+% 'dailyplanet' font theme uses same font family of "Fira Sans/Mono" that Metropolis 
+% default does, but works with heavier weight fonts for easier readability (small
+% screens in mind). Majority of font is in "Fira Sans Book".
+
+\ProvidesPackage{beamerfontthemedailyplanet}
+
+\RequirePackage{fontspec}
+
+
+\defaultfontfeatures{Mapping=tex-text}
+\setsansfont[BoldItalicFont={Fira Sans SemiBold Italic}, ItalicFont={Fira Sans Book Italic}, BoldFont={Fira Sans SemiBold}]{Fira Sans Book}
+\setmonofont{Fira Mono}
+\newfontfamily\Light{Fira Sans Light}
+\newfontfamily\Book{Fira Sans Book}
+\newfontfamily\Regular{Fira Sans}
+\newfontfamily\SemiBold{Fira Sans SemiBold}
+
+
+
+\AtBeginEnvironment{tabular}{\setsansfont[BoldFont=Fira Sans SemiBold, Numbers={Monospaced}]{Fira Sans Book}}
+
+
+\setbeamerfont{title}{family=\Regular, size=\Large}
+\setbeamerfont{author}{family=\Book, size=\normalsize}
+\setbeamerfont{date}{family=\Light, size=\normalsize}
+
+\setbeamerfont{section title}{family=\Regular, size=\Large}
+
+\setbeamerfont{block title}{family=\Regular, size=\normalsize}
+\setbeamerfont{block title alerted}{family=\Regular,size=\normalsize}
+
+\setbeamerfont{subtitle}{family=\Light, size=\fontsize{12}{14}}
+\setbeamerfont{frametitle}{family=\Regular, size=\large}
+
+\setbeamerfont{caption}{size=\small}
+\setbeamerfont{caption name}{family=\SemiBold}
+
+\setbeamerfont{description item}{family=\Book}
+
+\setbeamerfont{page number in head/foot}{size=\scriptsize}
+
+\setbeamerfont{bibliography entry author}{family=\Book, size=\normalsize}
+\setbeamerfont{bibliography entry title}{family=\SemiBold, size=\normalsize}
+\setbeamerfont{bibliography entry location}{family=\Light, size=\normalsize}
+\setbeamerfont{bibliography entry note}{family=\Light, size=\small}
+
+\linespread{1.15}


benjamin-weiss:
> I really struggled with getting a bibliography to work in Beamer using xelatex and ultimately gave up.

I used the following `frame` with no problems. Hope that helps. If you are still having problems I will make a screenshot for you.

    \begin{frame}{References}
        \begin{thebibliography}{10}
        
        \beamertemplatebookbibitems
        \bibitem{Oppenheim2009}
        Alan~V.~Oppenheim
        \newblock Discrete-Time Signal Processing
        \newblock Prentice Hall Press, 2009
        \newblock Notes nobody is interested in

        \beamertemplatearticlebibitems
        \bibitem{EBU2011}
        European~Broadcasting~Union
        \newblock Specification of the Broadcast Wave Format (BWF)
        \newblock 2011
      \end{thebibliography}
    \end{frame}

>  Is `dailyplanet` an acceptable name?

Looking at your screenshot the font theme looks very similar to `metropolis`. So I think it would be better to either choose a name like `metropolisBook` or (even better imho) implement it as an option for the original font theme.

> fira // Fira Sans Fonts: Provides Fira Sans and Fira Mono fonts, ...

It's probably better than nothing, but personally I don't understand why people cling to the old pdfTeX.

>  newtxsf // Sans Math Fonts: Provides sans-serif math fonts, such as Greek. Right now, I think there's a bit of an issue in metropolis with using Greek symbols, since they feel so different from Fira Sans.

As I wrote the [hsrmbeamertheme](https://github.com/benjamin-weiss/hsrmbeamertheme/) I looked a lot  into the problem of sans-serif math. And at the time, there was no satisfying solution. The main problem is, that there are very few fonts with full math support. So most of the time you end up replacing some symbols with the ones from a different font. And this always looks more or less strange. But it's probably worth using at least a package like `newtxsf`. Personally I gave up and decided to type-set my formulas in *Computer Modern Serif*. Better a clean break than a half-baked solution.

benjamin-weiss:
Another thought: I think before we start to add new font themes, we should revise the `metropolis` font theme with respect to minimizing the effort needed to write a new font theme respectively load a new font face. Ideally I think a new font "theme" should only contain two commands.

    \newfontfamily\main{⟨font name⟩}[⟨font features⟩]
    \newfontfamily\mono{⟨font name⟩}[⟨font features⟩]

But I have to think about that again and test some things if they work the way I remember it.

matze:
> Haven't touched the ReadMe yet. Wasn't sure if it was best to create another subsection (like the current Color customization), or list it somewhere else.

Concerning this issue and to avoid unnecessary work, I'd wait until #74 is merged.

ChipmunkMath:
**Bibliography:** Thanks @benjamin-weiss, that worked fine for me. I was trying really hard to use `biber` (for the first time), and now I feel a bit silly. Anyway, I think the new font theme looks good for bibliographies.

> Looking at your screenshot the font theme looks very similar to metropolis. So I think it would be better to either choose a name like `metropolisBook` or (even better imho) implement it as an option for the original font theme.

Yeah, it's super similar, just an increase in font-weight across the board. I'm fine with `metropolisBook` or similar, but I disagree with the idea of making it an option for the `metropolis` font theme. If you look at how Beamer deals with its stock font themes, it has the user call a new font theme for (I think) smaller overall changes (consider `structurebold`). Beyond matching how Beamer does things, I think it seems awkward for the user to call the default font theme just to pass an option to it. We could have the option given when calling the general theme (i.e., `\usetheme[bookfonts]{m}`), but the theme already has a pretty long options list.

>Another thought: I think before we start to add new font themes, we should revise the metropolis font theme with respect to minimizing the effort needed to write a new font theme respectively load a new font face. Ideally I think a new font "theme" should only contain two commands.

I agree with the spirit of this, but I don't think we can get it down to nearly as low a number as two commands. While this is only the first time I've really played with `fontspec` or changing fonts, I think a lot more manual choice is needed. While you *can* just set the main and mono fonts (`\setmainfont`, `\setmonofont`), that then causes `fontspec` to automatically assign bold, italic, bolditalic, etc. That may work okay for some fonts, but a font like Fira Sans has 15 different font-weights, so manual choice is important.

Furthermore, we have a lot of targeting for assigning fonts to specific Beamer sections (`frametitle`, `block title`, etc.), and that can only be semi-automated. It certainly would be possible to write a better template to work from. Something like this might be easier for creating new font themes:
```
\newfontfamily\Weak{<for low importance writing (side notes, etc)>}
\newfontfamily\Normal{<for most of the document>}
\newfontfamily\Strong{<for high importance writing (structure, etc)>}
\newfontfamily\BoldText{<proper bold text (bib titles, etc.)>}
```

>As I wrote the hsrmbeamertheme I looked a lot into the problem of sans-serif math. And at the time, there was no satisfying solution. The main problem is, that there are very few fonts with full math support.

Definitely. I was looking into it yesterday, and it's really hard to find a sans-serif choice for math. I was pretty pleased with `newtxsf`, though. I didn't test it extensively, but what I saw looked good. I'd recommend giving it a shot and seeing what you think. I'm trying to decide if I should use it myself, so if you do give it a try, I'd like to hear your opinion.

rchurchley:
I think the `fira` package is definitely worth investigating. The `pdflatex` compatibility would just be a free bonus — the package purports to provide Fira fonts for `xelatex` as well. This would remove the need for users to find and install the Fira fonts, which is the biggest barrier for a prospective user to adopt this theme. It would allow us to remove all the platform-specific instructions for installing the Fira fonts or manually changing the theme if Fira has an unexpected name.

> So I think it would be better to either choose a name like metropolisBook or (even better imho) implement it as an option for the original font theme.

I would actually suggest we change `beamerfontthememetropolis` to a more descriptive name like `beamerfontthemefira`. Once we're done #74 we should definitely revisit the font theme and add weight options.

ChipmunkMath:
>I think the fira package is definitely worth investigating. The pdflatex compatibility would just be a free bonus — the package purports to provide Fira fonts for xelatex as well. This would remove the need for users to find and install the Fira fonts, which is the biggest barrier for a prospective user to adopt this theme. It would allow us to remove all the platform-specific instructions for installing the Fira fonts or manually changing the theme if Fira has an unexpected name.

This is most of the reason I wanted to bring the `fira` package to everyone's attention. There is a potential issue in that `fira` is extremely recent, so unless a user has been doing a good job keeping their TeX install up to date, they might not have the package. Similarly, even if they have the package, it's important for them to be on the most recent version (extremely recent, less than a month ago), as the older version was missing some important things.  But it should be easier to tell a potential new user to just go run TeX Live (or equivalent) and make sure they have the most current version of `fira` than having them do the current install method.

As a caveat, though, I'd recommend extensive testing with the `fira` package before adopting it into the `metropolis` theme as a required package. When I was playing with it before, it didn't seem to be 100% perfect. It was very close, but there were some minor issues, so we should be careful before choosing to work with it.

rchurchley:
> Beyond matching how Beamer does things, I think it seems awkward for the user to call the default font theme just to pass an option to it. We could have the option given when calling the general theme (i.e., \usetheme[bookfonts]{m}), but the theme already has a pretty long options list.

I'm generally in favour of the option option — it goes against the beamer way but should be easier to maintain. I'm working on overhauling the options system using `pgfopts` and key-value pairs, which should make it much easier to pass options to the `metropolis` sub-packages.

benjamin-weiss:
> I'm fine with `metropolisBook` or similar, but I disagree with the idea of making it an option for the `metropolis` font theme. If you look at how Beamer deals with its stock font themes, it has the user call a new font theme for (I think) smaller overall changes (consider `structurebold`). Beyond matching how Beamer does things, I think it seems awkward for the user to call the default font theme just to pass an option to it. We could have the option given when calling the general theme (i.e., `\usetheme[bookfonts]{m}`), but the theme already has a pretty long options list.

Nevertheless `structurebold` also has options and of course the option would be a main theme option. Just like we do it with the `beamercolortheme.sty` options.

> It certainly would be possible to write a better template to work from. Something like this might be easier for creating new font themes:

Ok, I didn't see that you are actually using 4 different font weights and forgot that we are using 3 (thought it were only 2).  So I thought about it again. Maybe it's a good idea to adapt the CSS font weight naming scheme:

    \newfontfamily\normal{⟨font name⟩}[⟨font features⟩]
    \newfontfamily\bold{⟨font name⟩}[⟨font features⟩]
    \newfontfamily\bolder{⟨font name⟩}[⟨font features⟩]
    \newfontfamily\lighter{⟨font name⟩}[⟨font features⟩]

in addition to the monospaced font family.

    \newfontfamily\mono{⟨font name⟩}[⟨font features⟩]

Could that be the starting point for our font themes?

> the package purports to provide Fira fonts for `xelatex` as well.

I haven't seen that at first. Then we definitely should use it. 

> I would actually suggest we change `beamerfontthememetropolis` to a more descriptive name like `beamerfontthemefira`.

I'm on board with renaming to `beamerfontthemefira.sty`. 

>  I'm working on overhauling the options system using `pgfopts` and key-value pairs.

Good idea.

mlubin:
:+1:
I find the default font a bit too thin, especially for equations. This is great.

benjamin-weiss:
I think this PR is now obsolete after #145 has been merged.

@ChipmunkMath, @mlubin can you please check if a simple
```TeX
\setsansfont[BoldFont={Fira Sans SemiBold}]{Fira Sans Book}
```
after loading the theme satisfies your needs now?

ChipmunkMath:
Apologies for taking so ridiculously long to get back to you on this, @benjamin-weiss.

Putting `\setsansfont[BoldFont={Fira Sans SemiBold}]{Fira Sans Book}` in the preamble after loading the metropolis theme *mostly* works. For straight text in "normal" frames, it seems to work great. Here are the issues I've found:
- Does not affect the title page;
- Does not affect bibliography entries;
- Works pretty well on math, but doesn't necessarily affect everything. +, =, and probably many others are not made any thicker, but I think that's more an issue with math fonts than anything to do with Fira, right? My previous font theme had the same issue.
- Does not affect the `frametitle` (for me, a good thing: I like its current weight).

For me, none of the above are real issues, so I'm happy to use that as a fix. Other people might care about some of those mismatches, though.

Regardless of any further changes, I think some way to implement heavier weight characters should be included as an option in the package or at least the above `\setsansfont...` should be mentioned in the [manual](http://mirrors.ctan.org/macros/latex/contrib/beamer-contrib/themes/metropolis/doc/metropolistheme.pdf). While I like the aesthetics of the normal Fira Sans `Light`, I think it has issues with readability in a lot of situations.

If anybody wants to see them, here are sample images showing Fira Sans `Book` in action along with the title page and bibliography issues.
bookfontexamples-0
bookfontexamples-1

benjamin-weiss:
> Works pretty well on math, but doesn't necessarily affect everything. 

Yes, math is always a problem. The Fira Font is not a complete math font. So some of the characters are replaced. I personally prefer to typeset my math in plain old Computer Modern because of that.

> or at least the above `\setsansfont...` should be mentioned in the manual.

In section 3.3 the manual mentions how to change the fonts. Ok, it uses the Ubuntu font as an example, but from my point of view this is clear enough, but I am quite familiar with fontspec. So fell free to rephrase it and make a PR. 

About your other issues. Unfortunately I cannot reproduce any of them. Are you sure you are working at least with the CTAN version? It looks to me like you are still using an older version. Please check your texmf folder.

Here is my minimal example I tested with.

```TeX
\documentclass[10pt]{beamer}
\usetheme{metropolis}

\title{Metropolis}
\subtitle{A modern beamer theme}
\date{\today}
\author{Matthias Vogelgesang}
\institute{Center for modern beamer themes}

\begin{document}

\maketitle

\begin{frame}{Literaturverzeichnis}
  \begin{thebibliography}{10}
	\bibitem{Oppenheim2009}
	Alan~V.~Oppenheim
	\newblock Discrete-Time Signal Processing
	\newblock Prentice Hall Press, 2009
  \end{thebibliography}
\end{frame}

\setsansfont[BoldFont={Fira Sans SemiBold}]{Fira Sans Book}

\maketitle

\begin{frame}{Literaturverzeichnis}
  \begin{thebibliography}{10}
	\bibitem{Oppenheim2009}
	Alan~V.~Oppenheim
	\newblock Discrete-Time Signal Processing
	\newblock Prentice Hall Press, 2009
  \end{thebibliography}
\end{frame}

\end{document}
```

And the output I get.

boldtheme

ChipmunkMath:
>About your other issues. Unfortunately I cannot reproduce any of them. Are you sure you are working at least with the CTAN version? It looks to me like you are still using an older version. Please check your texmf folder.

Checking the log files after compiling with metropolis and I believe I am up to date (2015/12/04). I also see that version date when I manually look at the style files in my texmf directory.

Running @benjamin-weiss most recent MWE, both pairs of slides (lower font weight and higher font weight) are identical. Not sure what the differences in our setups could be.

rchurchley:
@ChipmunkMath, did my answer to #169 fix your problems here?

I can confirm that `\setsansfont[BoldFont={Fira Sans SemiBold}]{Fira Sans Book}` works as @benjamin-weiss describes, at least using the version of ᴍᴇᴛʀᴏᴘᴏʟɪs at the current `master` (b5910b8). I think it is probably safe to close this pull request for now; if @ChipmunkMath finds that it still doesn't work, we can reopen #169 or try to figure out the problem in a new issue.

ChipmunkMath:
@rchurchley Yes, it did! Thank you! (Sorry I didn't respond to #169: you posted your solution, but I didn't get the chance to test it before the issue was closed, so I wasn't sure if I should mention anything.)

Since I used the theme in the summer of 2015, my old files were getting in the way of things working properly. I cleaned them out, tried the MWE that @benjamin-weiss posted above (Jan 30), and everything works like a charm.

That said, I still think the theme's default font weight is a little too light for easy readability. While it's simple enough to put `\setsansfont[BoldFont={Fira Sans SemiBold}]{Fira Sans Book}` into the preamble of the document, I think something along these lines should be mentioned in the documentation. Having something about this in the "Font Customization" section of the documentation would be helpful for those users who want easier readability but aren't extremely TeX savvy.

rchurchley:
> Having something about this in the "Font Customization" section of the documentation would be helpful for those users who want easier readability but aren't extremely TeX savvy.

That's a good idea. Although we demonstrate `\usesansfont` in the manual, we should draw attention to the fact that this also works to increase the font weight. As you point out, there are presenting conditions (among them: large auditoriums, bright rooms, and small/underpowered projector) in which a heavier weight is necessary.

I have added a note in the documentation as part of pull request #195. Would you have anything to add to it, @ChipmunkMath?

PR #77, opened 2015-06-15
New demo slides

benjamin-weiss:
This PR adds a number of new slides to the demo presentation

- a TOC
- different block types
- a References slide


matze:
Can you recompile with
```latex
\renewcommand{\mthemetitleformat}{\scshape\MakeLowercase} 
```
set? I really can't stand camel case + small caps.

benjamin-weiss:
Sure, no problem.

I keep it in mind for the future. Or should we just add it to the `demo.tex`? If you are worried about complicating the `demo.tex` file, maybe we should prepare another example presentation. Like I did it with the `hsrmbeamertheme`: One big example presentation with a lot of dependencies and adjustments and another minimal example presentation.

matze:
> Or should we just add it to the demo.tex?

I think that's the way to go. As of now it's just a single line for which I don't think we need to split the demo presentation up.

benjamin-weiss:
Was a bit of a fight, to clean this PR up and rebase it properly, but I hope it is now ready to merge.

matze:
Thanks. To cheer you up: will get the hang of it :smile:

I don't know if you used that but for cleaning up I usually do interactive rebase on the branch (`git rebase -i HEAD~10`) which makes squashing, editing and removing commits a breeze.

benjamin-weiss:
I haven't heard of the interactive mode yet. I will look into it the next time I need a rebase. Actually at some point I did a `git reset --hard HEAD@{11}` because there where so many failed rebase, merges (still don't know where they came from) and so on. Basically I went so far back in time, that a simple `git rebase upstream/master` ran through without any conflicts. The changes I lost during that I then just recreated and combined them in a new commit. :laughing:

PR #79, opened 2015-06-15
Added custom block definitions, closes #76

### LINKED ISSUE #76: Block paddings and margins
The block paddings and margin do not look particularly nice when using `blockbg` (as can be seen on the first screenshot). What you see on the second screenshot is the definition I used in the hsrmbeamertheme. If you agree I will integrate the code into the theme once we decided how we will further organize the development regarding the `dtx` implementation. Maybe we can optimize it even more. E.g. a little `vskip` before the block or take a deeper look on how to adjust the block paddings and margins in case `blockbg` is *not* activated.

**Old**

screenshot_blocks_blockbg_old

**New**

screenshot_blocks_blockbg_new

benjamin-weiss:
As announced I optimized the old definitions I used in the `hsrmbeamertheme` a little bit. The height of the title block is a little bit reduced and the text in it properly vertically centered. Also the code is refactorized and should be easier to read. Just take a look at it if everything looks fine now.

screenshot_demo

screenshot_demo_blockbg

If so, then I will, as soon as PR #77 is merged, recompile the `demo.pdf` and rebase.

### DIFF
--- a/beamerthemem.dtx
+++ b/beamerthemem.dtx
@@ -378,6 +378,91 @@
 \setlength{\parskip}{0.5em}
 %    \end{macrocode}
 %
+% Block environment
+%
+%    \begin{macrocode}
+
+\newlength{\leftrightskip}
+\if@beamer@metropolis@blockbg
+  \setlength{\leftrightskip}{1ex}
+\else
+  \setlength{\leftrightskip}{0ex}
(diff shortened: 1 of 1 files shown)


matze:
Much, much better than the old layout. Thanks!

PR #83, opened 2015-06-16
moved missing information from README.md to mtheme.dtx

benjamin-weiss:
@matze Preferably first merge #81 and then I will rebase this PR.

Issue #82, opened 2015-06-16
Remove unused files


rchurchley:
I'm curious to see how many of the current files are unused and could be deleted.

- I assume `images/valley.jpg` was once used in the demo slides, but currently is not.
- If `make` is our supported build method, then `build.sh` and `latexmkrc` might be unnecessary.
- Do we need to complicate the project with `contributors.py`, `contributors.tex` and `login.json`? Personally, I think it's sufficient to acknowledge us contributors with a [link to the full list](https://github.com/matze/mtheme/graphs/contributors), which has the advantage of being automatically updated without the need for us to maintain a parallel list.
- We should investigate the `fira` bundle to see if it can be relied on to provide the fonts. If so, we will be able to remove `contrib/Dockerfile` and `contrib/getFiraFont.sh`.

matze:
> I assume images/valley.jpg was once used in the demo slides, but currently is not.

Yes, that's a leftover that was used to demonstrate the `\plain` command.

> build.sh and latexmkrc might be unnecessary.

I remember that `build.sh` was called in the Dockerfile. But apparently that's not the case anymore, so we can safely remove that. Concerning `latexmkrc`, yes we should remove that too and prominently advertise Make-based builds.

> We should investigate the fira bundle to see if it can be relied on to provide the fonts.

What is the `fira` bundle?

benjamin-weiss:
> Do we need to complicate the project with `contributors.py`, `contributors.tex` and `login.json`?

Yeah it probably got a little out of hand. But I kind of like it to have the complete list inside the documentation. But your way is also ok with me.

> What is the `fira` bundle?

A fairly new package which provides the Fira Sans font face. See the discussion in #75.

benjamin-weiss:
The contributors issue is resolved. It seems like we decided to keep the docker related files for now. So it's only left to remove the images folder and we can close this, right?

matze:
And the `latexmkrc` but since there are no objections I will just remove both.

PR #84, opened 2015-06-16
optimized Makefile structure and improved install/uninstall

benjamin-weiss:
As proposed by @rchurchley in #82 I also removed the contributors
script. The corresponding changes in the mtheme.dtx are already
implemented in #83.

The additional make rules are already described in the documentation
(also #83)


benjamin-weiss:
Should be a safe uninstall now.

@matze By the way, do you still want to keep the docker build rules for a little bit? Or shall I remove them right now within this PR?

matze:
I will redirect that question to @awalterschulze. I personally don't use Docker but if there is a large user base that benefits from this, I'd keep it.

awalterschulze:
I am a user :)
It is quite nice to be able to git clone and then start editing the example file and just run make docker-run to get a pdf.
Maybe its just me, but in the worst case these are very nice and testable installation instructions.

On a side note: The Dockerfile is in the contrib folder, but it should be in the root folder.

rchurchley:
As far as I can see, the main purpose of the Dockerfile is to allow some users an easy way of installing the Fira fonts. I think it can and probably should stay until we can move to a different solution for font installation (see #85)

awalterschulze:
Basically yes, but also its a nice way to test the font installation.

benjamin-weiss:
> I think it can and probably should stay until we can move to a different solution for font installation (see #85).

Ok, so let's keep it for now.

benjamin-weiss:
As it seems we decided to keep the docker build rules, I think this one can also be merged.

matze:
Alright, thanks!

PR #88, opened 2015-06-17
Split off and document pgfplots theme

rchurchley:
This pull request addresses #69 and adds some documentation to the `pgfplots` theme. The new theme is called `pgfplotsthemetol`, since the colours are based off of Paul Tol's palettes. The main theme only loads the sub-package if the user loads `pgfplots` themselves in their preamble. (Thanks `etoolbox` for providing `\AtEndPreamble` to make this work!)


matze:
I guess we are (or will be) running into the separation of concern issues sooner or later. Right now the pgfplot theme is tied to a particular color. Ideally we would have a pgf color theme, which could actually come from the color theme itself in order to provide consistency, and a structural theme, i.e. all those axes, legend and grid customizations. But maybe we should tackle that in a separate issue/PR.

benjamin-weiss:
@rchurchley You forgot to adjust the following lines:

    %<driver> \usepackage{beamerthemem}
    %<driver> \begin{document}
    %<driver> \DocInput{beamerthemem.dtx}
    %<driver> \end{document}

benjamin-weiss:
> But maybe we should tackle that in a separate issue/PR.

Yes in #69 I suppose?

@rchurchley Shouldn't we move the introduction comment into the `mtheme.dtx`. I think that's valuable information for the ordinary user.

rchurchley:
> Ideally we would have a pgf color theme, which could actually come from the color theme itself in order to provide consistency, and a structural theme, i.e. all those axes, legend and grid customizations.

Since the pgfplots colour theme uses far more distinct colours than any beamer theme would, I don't think it would be practical to define these colours in terms of the beamer theme. I'm not sure whether the colour theme is separable from the structural one, either: since the styles are not applied automatically, I think splitting them up would also require the user to apply two different styles to each chart they create. But I don't really use `pgfplots`, so I could be wrong there.

(The current implementation also has the advantage of being usable outside of beamer; in fact, it arguably could be its own package independent of this theme.)

> By the way, great work!

Thanks!

>  You forgot to adjust the following lines

Oops! Now fixed.

> Shouldn't we move the introduction comment into the mtheme.dtx. I think that's valuable information for the ordinary user.

Good idea — done and done.

benjamin-weiss:
> The current implementation also has the advantage of being usable outside of beamer; in fact, it arguably could be its own package independent of this theme.

Maybe we should really do that. Actually that was my first though when I read the SRON paper for the first time a while ago: "It would be really nice to have all the `color`, `plotcyclelist` and `colormaps` definitions in a nice little LaTeX package."

> By the way, great work!

+1

rchurchley:
Shall we then merge now to separate the `pgfplots` theme from the main theme and figure out what to do with it in #69 and a new PR?

matze:
Sounds like a plan!

Issue #89, opened 2015-06-17
Split `beamerthemem` into inner and outer themes


rchurchley:
I suggest we break `beamerthemem.sty` into an "inner theme" and "outer theme" similar to the structure of `beamerthemedefault`. The main theme file would then include the various sub-packages and forward options appropriately. This would allow users to apply the structural parts of the theme independently from the colour and font themes. This would address outstanding issues with #8 and #85.

matze:
Fixed by #93.

Issue #90, opened 2015-06-17
Use key-value options


rchurchley:
The `pgfopts` package allows for options of the form `key=value`, which would let us reimplement the options list with much more memorable names. For example:

```
noslidenumbers & usetotalslideindicator
    -> numbering={none, count, fraction}

protectframetitle & \mthemetitleformat{} 
    -> frametitle={plain, protect, lowercase, uppercase}

darkcolors
    -> colors={light, dark}
```

matze:
:+1: on all accounts.

rchurchley:
I made an ill-advised attempt to implement this during pull request #93, which should have addressed #89 only. It's now been decided to deal with that issue properly in a future separate pull request, so discussion should revert back to here.

It turns out `pgfkeys` is very powerful but programming it can be a bit overwhelming. Hopefully, we can set up a bit of helper code so the interface for adding new options and changing old ones is as simple as possible. @benjamin-weiss suggested that this could (or should) be done through custom `pgfkeys` handlers. With the right definitions of `.choose between` and `.compare`, it should theoretically be possible to have something like this:
```latex
\usepackage{beamerthememetropolisfunctions} % contains helper code
\pgfkeys{
  /metropolis/.cd,
  numbering/.choose between={none, counter, fraction},
  numbering=counter, % set default value
}

\pgfkeys{/metropolis/numbering/.compare=counter} % returns "true"
\pgfkeys{/metropolis/numbering/.compare=fraction} % returns "false"

\pgfkeys{/metropolis/numbering=none} % change keys mid-document
\pgfkeys{/metropolis/numbering=banana} % ...but throws a warning for bad values
```

I've managed to construct [a working example](https://gist.github.com/rchurchley/9ce4389bb39cdb90015a) of helper code that implements this. Comments?

benjamin-weiss:
Nice work on the handlers. It definitely seems to work. One minor remark: It should be an error and not only a warning in my opinion. By the way have you tested if your implementation is capable of setting multiple values at once?

Yet I'm still on the fence what implementation we should use. Yours definitely produces shorter option definitions. But what are we gonna do with the `section slide` like options? There we still would have to go back to an option definition like in my [Implementation 3](https://gist.github.com/benjamin-weiss/b11749b4af26f2e3aeac), right? So this would lead to two different kinds of option definitions. It would be nice to have one common style. 

[...]

PR #93, opened 2015-06-18
Split off and document inner and outer themes

rchurchley:
This is the big reorganization discussed in #89 and in other issues. It's not quite ready to be merged yet; I still have to

- tidy up the skeleton class `beamerthemem.dtx`,
- add a bit more documentation to `mtheme.dtx`, and
- reorganize the theme options so the inner and outer themes can be called independently of `beamerthemem.dtx`.

I have tried to keep functional changes to a minimum. The most notable one I had to make was to `\progressbar`, which has once again been broken up into two implementations, this time as custom beamer templates. The section page progress bar is defined in the inner theme as `progress bar in section page`, and the frametitle progress bar is defined in the outer theme as `progress bar in head/foot`. 

Although it is a bit annoying to have to split up the implementations because of the inner/outer theme distinction, we do get a side benefit from it. The colours of the two kinds of progress bar can now be changed independently:

```latex
\setbeamercolor{progress bar in section page}{...}
\setbeamercolor{progress bar in head/foot}{...}
```

The current functionality remains where `\setbeamercolor{progress bar}{...}` will change both, although doing so will not override any changes made using the more specific declarations above.

Since I'll be busy tomorrow and may not be able to finish this until the weekend, I thought I should make the pull request in the meantime so you all could start reviewing it. Apart from the known issues above, can you see anything I've missed or broken?


matze:
First of all: awesome work :+1:

> Although it is a bit annoying to have to split up the implementations because of the inner/outer theme distinction, we do get a side benefit from it. The colours of the two kinds of progress bar can now be changed independently

I like that, makes it easy to switch back to the non-alert progress bar in section slides.

> Apart from the known issues above, can you see anything I've missed or broken?

I tried different options and so far there are no obvious regressions, everything looks as it's supposed to be.

benjamin-weiss:
:+1: Really good work again.

>  The most notable one I had to make was to `\progressbar`, which has once again been broken up into two implementations ...

I suspected that as you first proposed the splitting. But as it gives us a plus on functionality, it is not really a downside.

Regarding missing or broken functionality: I couldn't spot any problems either.

benjamin-weiss:
As `beamerthemem.dtx` is now a very small file, shall we combine it with `mtheme.dtx`? It seems to be customary to have only one `dtx` file with the main package and the documentation. This should probably also solve the `\GetFileInfo{•}` issue.

rchurchley:
I should apologize here for pushing two commits that really ought to have been their own pull request. I got to work reorganizing the options, one thing led to another, and now I've implemented #90. I will do more rigorous testing tomorrow — followed by quite a bit doc-writing — but in the meantime let me know if you find any problems or have suggestions about the key-value option names.

> As `beamerthemem.dtx` is now a very small file, shall we combine it with `mtheme.dtx`?

I would not object to this. The only thing to consider is whether it will be more or less convenient to update the options list at the top of `beamerthemem.dtx` whenever a new one is added. But since the documentation also has to be updated any time that happens anyways, I don't foresee any major problems.

benjamin-weiss:
Your `pgfopts` implementation doesn't feel very nice to read. E.g. I haven't yet understood where you set the default values. And it has a very important downside. If you pass an illegal key value it just gets ignored and you don't get an error. E.g.

    \usetheme[numbering=fractin]{m}

So here are two proposals how the options can be implemented differently using `pgfopts`.

[Implementation 1](https://gist.github.com/benjamin-weiss/dbd350e0ce7872d4cc70)
[Implementation 2](https://gist.github.com/benjamin-weiss/c558045a6340f4cc9f6c)

The first one is similar to your implementation as it also uses `if` conditions, but it also produces a lot of code. The second is more straight forward, but is not very nice if the option has to insert a lot of code. Although the option code can always be a macro of its own.

Personally I think we should consider implementation 2 for most of the options. For some of the options (`sectionslide`, `blockbg`) I would still use a simple `true`/`false` switch. Look at the following gist for an example.

[Implementation 3](https://gist.github.com/benjamin-weiss/bac6b9a3da788829f9e6)

rchurchley:
> Your `pgfopts` implementation doesn't feel very nice to read. 

I agree. I hope we can come up with a good solution after a couple more iterations!

> And it has a very important downside. If you pass an illegal key value it just gets ignored and you don't get an error.

Good point. As in your first implementation shows, we'll have to use a `is choice` key somewhere to catch these errors. How about the following variant? We add `default` as a possible value for each key, like so:

```latex
\provideboolean{numbering/default}
% ...create other booleans named like `key/value`

\pgfkeys{% set up `numbering` key and possible values
  /metropolis/.cd,
  numbering/.is choice,
  numbering/default/.is if=numbering/default,
  numbering/none/.is if=numbering/none,
  numbering/counter/.is if=numbering/counter,
  numbering/fraction/.is if=numbering/fraction,
  numbering=default
}
```
Then, when the user calls `numbering=value` we only have to reset the `numbering/default` boolean. (This assumes the user only provides one value for each key, but I think we can trust the user to be reasonable here; it may also be beneficial in the future to allow keys to take on multiple values simultaneously.)
```latex
\pgfkeys{% define what happens when the user calls numbering=value
  /metropolis/options/.cd,
  numbering/.code={%
    \pgfkeys{/metropolis/numbering/default=false}
    \pgfkeys{/metropolis/numbering=#1}
  }
}
\ProcessPgfOptions{/metropolis/options}
```
Having the `default` option will not overly complicate the rest of the code, as we can use `etoolbox`'s boolean expressions:
```latex
\ifboolexpr{numbering/default or numbering/counter}{%
    \insertframenumber%
}{}
```
This stuff is complicated enough that it might be worth providing some macros, perhaps in a `beamerthememetropolisfunctions.sty` file, for setting options. I managed to come up with this:
```latex
\providecommand{\metropolis@defineoption}[1]{
  \pgfkeys{/metropolis/#1/.is choice}
  \metropolis@optionvalue{#1}{default}
  \pgfkeys{/metropolis/#1=default}
  \pgfkeys{/metropolis/options/#1/.code={%
    \pgfkeys{/metropolis/#1/default=false}
    \pgfkeys{/metropolis/#1=##1}}
  }
}

\providecommand{\metropolis@defineoptionvalue}[2]{
  \providebool{#1/#2}
  \pgfkeys{/metropolis/#1/#2/.is if=#1/#2}
}
```
This makes the start of each package more readable:
```latex
\RequirePackage{beamerthememetropolisfunctions}
\metropolis@defineoption{color}
\metropolis@defineoptionvalue{color}{light}
\metropolis@defineoptionvalue{color}{dark}
\metropolis@defineoption{block}
\metropolis@defineoptionvalue{block}{none}
\metropolis@defineoptionvalue{block}{fill}
```
What do you think?

> For some of the options (sectionslide, blockbg) I would still use a simple true/false switch.

I deliberately chose those options to not be true/false to leave open the possibility of offering other styles. For example, `sectionslide=plain` could use a plain line like the title page instead of a progress bar, or somebody might ask find a reason to ask for `block=outline`.

benjamin-weiss:
> I agree. I hope we can come up with a good solution after a couple more iterations!

We most definitely will. I would really prefer to exchange MWE like my implementations 1/2 for this. So, if you already have it lying around maybe you could set up a gist for it. But either way I will have a deeper look at your new proposal.

> I deliberately chose those options to not be true/false to leave open the possibility of offering other styles. 

Ok. Makes sense. I only thought about the current situation and not about the future.

rchurchley:
> I would really prefer to exchange MWE like my implementations 1/2 for this. So, if you already have it lying around maybe you could set up a gist for it.

Sorry about that! [Here's the MWE I've been testing with.](https://gist.github.com/rchurchley/9ce4389bb39cdb90015a)

benjamin-weiss:
Thank you. Saves a lot of time. Just tried to set up your proposal and just got error messages at first.

benjamin-weiss:
As you are also seem to be working on it right now I just write things down as soon as I discover them.

If you change the value mid-document, the bool values are wrong. Meaning you get two outputs.

[MWE](https://gist.github.com/benjamin-weiss/a488619633613b34cbb4)

Another problem are imho your `\metropolis@define` macros in the first place. When things get too complicated we probably should define our own handlers instead of macros.

rchurchley:
> If you change the value mid-document, the bool values are wrong. Meaning you get two outputs.

This is definitely correct. It is an open question whether we would like to support this — we currently don't, and we'd have to rework the structure of a few packages to make it work even if the truth values were correct. But using `pgfkeys` makes mid-document option switching theoretically possible. I'll look into defining handlers and see what they can do for us.

benjamin-weiss:
> But using pgfkeys makes mid-document option switching theoretically possible.

And I would really love to support it.

benjamin-weiss:
Just thinking out loud here. The following code demonstrates how I expect the setting up of options should work, once we have written our own handler (`is ifchoice`).

```TeX
\newif\ifnumbering@none
\newif\ifnumbering@counter
\newif\ifnumbering@fraction

\pgfkeys{
  /metropolis/.cd,
  numbering/.is ifchoice,  % new handler ifchoice
  numbering=counter, % set default value
  numbering/none/.is if=numbering@none,
  numbering/counter/.is if=numbering@counter,
  numbering/fraction/.is if=numbering@fraction,
}

% Should set all the ifs accordingly
\pgfkeys{/metropolis/numbering=fraction}

\ProcessPgfOptions{/metropolis}

% use the if conditions whenever you want and get the right result
\ifnumbering@none
  -
\fi
\ifnumbering@counter
  1
\fi
\ifnumbering@counter
  1/9
\fi
```

rchurchley:
I think that would be ideal. The trick is: is it possible to somehow to get a list of subkeys of `numbering`? To implement this, we need `numbering/.is ifchoice` to loop over `numbering/*` and set their values all false before `\pgfkeysalso\expandafter{\pgfkeyscurrentkey/##1}`. Do you know if there's a way to do this?

benjamin-weiss:
And for things like the section slide the following pseudocode should be the goal.

```TeX

\pgfkeys{
  /metropolis/.cd,
  section slide/.is choice,  % new handler ifchoice
  section slide=progress bar, % set default value
  section slide/none/.code={\sectionslide@none},
  section slide/progress bar/.code={\sectionslide@progressbar},
  section slide/plain/.code={\sectionslide@plain},
}

\ProcessPgfOptions{/metropolis}

% set up macros for each choice
\newcommand{\sectionslide@none}{%
  \AtBeginSection{
    % intenionally empty
  }
}

\newcommand{\sectionslide@progressbar}{%
  \AtBeginSection{
    % code for sectionpage with progressbar
  }
}

\newcommand{\sectionslide@plain}{%
  \AtBeginSection{
    % code for sectionpage with simple line
  }
}
```

> Do you know if there's a way to do this?

Not yet. :smirk: I'm still new to `pgfkeys` from the programmer side. We should probably have a look at the original `.if choice` code. If you find it tell me. `pgf` is such a big package that it is kind of hard to find the piece of code you are looking for.

rchurchley:
Just looking at it right now. You can find it on your own system with `kpsewhich pgfkeys.code.tex`:

```latex
\pgfkeys{/handlers/.is choice/.code=%
  \pgfkeys{%
    \pgfkeyscurrentpath/.cd,%
    .code=\def\pgfkeys@was@choice{##1}\expandafter\pgfkeysalso\expandafter{\pgfkeyscurrentkey/##1},
    .unknown/.code={%
      \def\pgf@marshal{\pgfkeysvalueof{/errors/unknown choice value/.@cmd}}%
      {\expandafter\expandafter\expandafter\pgf@marshal\expandafter\expandafter\expandafter{\expandafter\the\expandafter\pgfkeys@pathtoks\expandafter}\expandafter{\pgfkeys@was@choice}\pgfeov}%
    }%
  }%
}
```
I haven't yet found a structure that would give us something we could loop over.

> Not yet. :smirk: I'm still new to pgfkeys from the programmer side.

I'm actually new to `pgfkeys` on the user side, so I appreciate all the hints you've given so far about what's possible!

benjamin-weiss:
Making the option definition even simpler, like this

```TeX
\newif\ifnumbering@none
\newif\ifnumbering@counter
\newif\ifnumbering@fraction

\pgfkeys{
  /metropolis/.cd,
  % new handler ifchoice
  numbering/.is ifchoice={numbering@none, numbering@counter, numbering@fraction},
  numbering= numbering@counter, % set default value
}
```

should probably give us the opportunity to save the list for later.

benjamin-weiss:
After thinking about it again: the list should be a list of key values.

```TeX
numbering/.is ifchoice={%
  none=numbering@none,
  counter=numbering@counter,
  fraction=numbering@fraction,
},
```

rchurchley:
We might be able to simplify further and not have to maintain all the `if`s by making the key store its result and figuring out a `.compare` handler:

```
\pgfkeys{
  /metropolis/.cd,
  % new handler ifchoice
  numbering/.is storechoice={none, counter, fraction},
  numbering=counter, % set default value
}

\pgfkeys{/metropolis/numbering/.compare=counter} % returns "true"
\pgfkeys{/metropolis/numbering/.compare=fraction} % returns "false"
```

rchurchley:
Oooh, it may even be possible to do the same thing with the stock `.is choice`:
```latex
\pgfkeys{
  /metropolis/.cd,
  numbering/.is choice,
  numbering/none/.value, % new handler .value
  numbering/counter/.value,
  numbering/fraction/.value,
  numbering=counter % executes /counter/.code, which .value defines to \pgfkeyssetvalue{numbering}{counter}
}

\pgfkeys{/metropolis/numbering/.compare=counter} % new handler compare, returns "true"
\pgfkeys{/metropolis/numbering/.compare=fraction} % returns "false"

**Comment by benjamin-weiss** [comment]
I like the idea, but it would produce quite long if statements. Yet this would be acceptable. Even if it is possible with the stock `.is choice` I prefer your first proposal. It is more work to get it running in the first place but produces nicer option definitions.

**Comment by rchurchley** [comment]
I'll try to see what I can do to get a MWE tomorrow then!

**Comment by benjamin-weiss** [comment]
Alright, I will also keep looking into it. So in case I have a MWE I will post it. Just to be clear: Our goal is to implement your first "`store choice`" proposal?

**Comment by rchurchley** [comment]
Yes, I agree that my first proposal is nicer than the second. But after further thought, the best of both worlds should be possible by setting up a handler to do the repetitive bits of my second proposal.
```
\pgfkeys{
  /metropolis/.cd,
  numbering/.is choice,
  numering/.values={none, counter, fraction},   % new handler .values
  numbering=counter, % set default value
}

\pgfkeys{/metropolis/numbering/.compare=counter} % new handler .compare, returns "true"
\pgfkeys{/metropolis/numbering/.compare=fraction} % returns "false"
```

benjamin-weiss:
It don't get why you think an additional `.value` handler is necessary.

rchurchley:
I suppose you're right — you only need two. The `.values` handler can set up the code for the value keys when it sets them up, without having to go through an intermediate handler. Such are the dangers of thinking (and coding) aloud!

rchurchley:
Final thought before I go to bed — it finally dawned on me that a custom handler `.is storechoice` (or whatever) doesn't have to copy code from the base `.is choice`, it can just call it. In that case I have no preference between the first proposal (with `.is storechoice`) and the third (with `.values`).

benjamin-weiss:
Trying to implement a custom handler drove me insane. `pgfkeys` just doesn't do what I want. :smirk:

So I improved my implementation 2 a little bit. The implementation provides every functionality I expect and is from my point of view easy to understand and to implement.

[Implementation 3](https://gist.github.com/benjamin-weiss/b11749b4af26f2e3aeac)

So, what do you think about it?

benjamin-weiss:
@rchurchley So? Where you able to implement a working custom handler?

Another thing: What are we gonna do with your last two commits. AFAICT nearly all the changes you made are related to `pgfotps`, except of the documentation you added. So don't you think we should (soft) revert them and just merge this PR. As the file splitting is a major change I would like to see it integrated in the master branch asap.

rchurchley:
> What are we gonna do with your last two commits. AFAICT nearly all the changes you made are related to pgfotps, except of the documentation you added. So don't you think we should (soft) revert them and just merge this PR. As the file splitting is a major change I would like to see it integrated in the master branch asap.

Many apologies! As this long diversion has shown, I was wrong to try to bring #90 into this.  Lesson learned — I'll try to keep to one major feature per PR. (I hope) I've got this one back to a mergeable state.

rchurchley:
As for custom handlers, I think I've got something working! See [this updated gist](https://gist.github.com/rchurchley/9ce4389bb39cdb90015a). I'll cross-post that to #90, where we can move this branch of the conversation.

matze:
As far as I can see all commits here are now related to this PR's title? That means I can merge?

benjamin-weiss:
I think so. @rchurchley ?

rchurchley:
The inner and outer themes can't be called independently of the main theme yet until we rework the options, but we can finish that in #90. I think it should be fine to merge.

PR #94, opened 2015-06-18
Demonstrate the [allowframebreaks] works by default

### LINKED ISSUE #20: "TeX capacity exceeded" when using allowframebreaks

elliottd:
Fixes #20

### DIFF
--- a/beamerthemem.dtx
+++ b/beamerthemem.dtx
@@ -511,7 +511,7 @@
 \if@protectFrameTitle%
     \mthemetitleformat{\protect\insertframetitle}%
 \else%
-    \mthemetitleformat{\protect\insertframetitle}%
+    \mthemetitleformat{\insertframetitle}%
 \fi%
 \end{beamercolorbox}%
 \if@useTitleProgressBar
--- a/beamerthemem.dtx
+++ b/beamerthemem.dtx
@@ -511,7 +511,7 @@
 \if@protectFrameTitle%
     \mthemetitleformat{\protect\insertframetitle}%
 \else%
-    \mthemetitleformat{\insertframetitle}%
+    \mthemetitleformat{\protect\insertframetitle}%
 \fi%
 \end{beamercolorbox}%
 \if@useTitleProgressBar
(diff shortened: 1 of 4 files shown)


matze:
You get the error even when using `\usetheme[protectframetitle]{m}`? That's in essence what the condition is doing there.

elliottd:
Oh, I didn't notice that was an option for the theme package because I
didn't read the PDF.

Would there still be value in merging demo.{tex, bib} and the new build.sh to
demonstrate that allowframebreaks works correctly?

benjamin-weiss:
I don't think @matze will like that, as it forces the `demo.pdf` to use camel case titles. What about that: Let's create a second `demo.tex` to demonstrate solely the use of `protectframetitle`?

elliottd:
I understand what you're saying about tge camelCasing. However, this is a
beamer theme, where many of the intended users will want to include a
bibliography.

The default behaviour of copying demo.tex into your own presentation is
that it won't insert all of the references. The next step will likely be to
apply [allowframebreaks] to the references slide, and then the document
won't compile.

I'm not sure that's the experience we want people to have with this theme.

benjamin-weiss:
@elliottd Can you please have a look at my PR #96 and tell me if it solves the issue for real now and that it doesn't have any unwanted side effects?

elliottd:
@benjamin-weiss I don't know enough TeX to provide an answer for the second part of your question. Perhaps we should wait for someone with more experience to discuss whether it has any unwanted side effects.

benjamin-weiss:
I just meant to test it from the user side. So just typeset your presentations with the new implementation and check whether it solves the problem as you expect it as a user and if the patch is breaking any functionality. In principle the implementation does work with no side effects at all.

rchurchley:
I think this PR can safely be closed, since the changes were incorporated into #96. Thanks again for the suggestion!

PR #96, opened 2015-06-20
New frame title format implementation

benjamin-weiss:
As also suggested on [stackexchange](http://tex.stackexchange.com/questions/112526/errors-wit
h-makeuppercase-and-beamers-insertsectionhead) the new implementation patches `\sectionentry`, `\beamer@section` and `\beamer@@frametitle` instead of expanding them.

In the process I also added two more hooks:

- `mthemesectiontitleformat`
- `mthemeframetitleformat`

So the user can change the behavior of each element independently. As default the new formats are dependent on `mthemetitleformat`. Maybe we should even add a third one for the plain frames.

For testing purposes I integrated @elliottd s frames from his PR #94

This should fix the `allowframebreaks` issue for real now.


benjamin-weiss:
When #90 is implemented we should maybe make the case setting an option and the small caps setting should either way be done 'beamer like' via

    \setbeamerfont{frametitle}{shape=\scshape}

and so on.

benjamin-weiss:
@matze Maybe you can have a quick look at it, that I didn't break anything. But I think this PR should be ready to merge.

matze:
Looks reasonable!

Issue #95, opened 2015-06-18
Minor design improvements for the title page


benjamin-weiss:
Even if the title page looks well designed on its own, it feels a little bit too light in comparison to the other slides. First of all it is the only place we use the `ExtraLight` font (for the author and the date) and the dividing line is a little bit thinner than the progress bar.

So I think we should change the author and date font to `Light`, change the institute font to `ExtraLight` and use the same thickness for the divider and the progress bar.

![screenshot_titlepage](https://cloud.githubusercontent.com/assets/9158719/8244614/4246f14c-1623-11e5-81e0-be471f9f5797.png)

If you guys agree I will make a PR.

matze:
*Going to see the difference …* yes, go for it. By the way, maybe we could include a fake logo or use one of the symbols on slide 21 to represent something like an institutional logo. I guess it's not too uncommon to have some such on the very first slide.

ChipmunkMath:
As someone who needs a logo on their title page, I would love to see some sort of easy image inclusion.

benjamin-weiss:
I don't really like the idea of a "random" logo. But an official mtheme logo would be nice. Wanna design one? Sadly my design skills would produce a logo unworthy of the theme. :smile:

> I would love to see some sort of easy image inclusion.

That is already possible. See #30. Maybe we should mention that in the documentation.

matze:
Well, considering that the theme's (still?) inofficial name is Metropolis, we could use the logo from the iconic movie

![Metropolis logo](https://upload.wikimedia.org/wikipedia/commons/thumb/8/83/Metropolis1927-logo.svg/200px-Metropolis1927-logo.svg.png)

but that might be a bit too extreme. I'll try to come up with something.

benjamin-weiss:
Hehe. Interesting idea. But this logo basically screams *noise* and not so much *minimal visual noise*. So it is not really fitting I think.

bgat:
Related, at some professional conferences the logo appears on the lower
right of the title page, and as a banner in the footers of body pages. I'm
not against hacking the template myself to accomplish this, but haven't
been able to do so thus far without wrecking the whole thing. Is there a
structural improvement that would make this easier?

Second, I prefer to include a "title page" at the end of the document,
which is just like the first page but adds the copyright/redistribution
terms for the material in small print near the bottom of the frame. Can you
consider adding a \license{} variable, and then e.g. a parameter to
\maketitle that will show that variable's contents if requested?

[...]

PR #100, opened 2015-06-23
upper part of logo redrawn by hand to get proper transparency

benjamin-weiss:
I have redrawn your "vectorized chaos" :smirk: in the top part of the logo to get a proper transparency. I hope it still looks like you intended.

Kind of strange is the size of the `eps` and `pdf` that Illustrator produces: `svg` → 8 KB `eps` → 986 KB

Another thing: It would be nice if we could set the logo color directly in LaTeX. Any idea whats the best way to do that?


rchurchley:
Opening that `svg` in Inkscape and saving it as an `eps` results in a 21kB file. I think the `svg` filesize can be reduced as well if you put it through [`svgo`](https://github.com/svg/svgo).

benjamin-weiss:
Thanks for the info. I thought about trying to convert it with inkscape. But I was too lazy to install it. :smirk:

rchurchley:
> It would be nice if we could set the logo color directly in LaTeX. Any idea whats the best way to do that?

It turns out that TikZ includes an operation to draws a path specified in the SVG path data syntax. The TikZ manual gives an example:
```latex
\begin{tikzpicture}
  \filldraw [fill=red!20] (0,1) svg[scale=2] {h 10 v 10 h -10}
    node [above left] {upper left} -- cycle;
  \draw svg {M 0 0 L 20 20 h 10 a 10 10 0 0 0 -20 0};
\end{tikzpicture}
```
Using this, it should not be too difficult to express the paths in `logo/logo-redrawn.svg` to TikZ, allowing us to automatically take colours from the theme.

matze:
@benjamin-weiss: you can just overwrite the original file(s). I just wanted to get #95 get going :smile:

Concerning re-doing the logo, I am sceptical because than we cannot re-use it easily for other purposes. But if we decide, to just keep it within the theme itself, why not.

benjamin-weiss:
> you can just overwrite the original file(s)

I will do that. Just wanted to solve the file size issue first.

> It turns out that TikZ includes an operation to draws a path specified in the SVG path data syntax.

Nice. It seems there is even an [export extension for inkscape](https://github.com/kjellmf/svg2tikz).

rchurchley:
I think it's fine to go with just the `svg`-`eps`-`pdf` options — I don't expect the logo to show up anywhere but the demo slides, so it doesn't necessarily need to change colour. Just thought I'd mention that taking colours from the scheme was possible :)

I would like to cut down on the file sizes before merging this, however. Can you re-export with Inkscape, @benjamin-weiss, or would you prefer one of us to do it?

benjamin-weiss:
> Can you re-export with Inkscape, @benjamin-weiss, or would you prefer one of us to do it?

I will do it. Even though we probably not use it. I'm curios about the TikZ exporter. So I will eventually install inkscape after all.

rchurchley:
I finally realized that this pull request was updated eight months ago with much better file sizes! Thanks for doing that, and sorry it took us so long to notice.

I'm still not sure we need to keep around all 995kB of the full Illustrator-generated file. If edibility is a concern, then we can keep around a second Inkscape-generated file where the text is not converted to paths (roughly 16kB).

Apart from that, I think this is a good tweak!

benjamin-weiss:
I rebased this PR. Do we want to keep a version where the text is not converted to paths? I removed it for the time being, but still have it (as a small version saved with inkscape).

rchurchley:
Perfect. Looks like we can merge now!

I don't think we need to keep the non-converted version as a separate file in the repo. However, you could add that version and then revert it — that will keep the demo directory clean while ensuring we still have the text-editable file still available in the commit history.

Issue #102, opened 2015-06-25
`beamerthemem.sty' not found.


digen:
Hi there, 
Thanks for the theme. Facing a bit of an issue getting it to build inside a docker container. 

When I run the build.sh, get the following,
```
! LaTeX Error: File `beamerthemem.sty' not found.

Type X to quit or <RETURN> to proceed,
or enter new name. (Default extension: sty)

Enter file name:
```

The file beamerthemem.sty does not exist in the current working directory(repo). Surprisingly, the repo which I had downloaded a few days ago on a different computer has the file. Can anyone point out if I am missing some step ?

benjamin-weiss:
Try to run `make sty` before building the docker container. We converted the theme to the "Documented LaTeX Format", so the `sty` files must be created first.

@eddelbuettel, @awalterschulze I personally do not use docker, so I can't really help fixing the build rules. Can you have a look at the docker build rules again.

digen:
@benjamin-weiss That fixed it. Thanks !

awalterschulze:
I'll take a look at this.

eddelbuettel:
I went the other way and just created a .deb for my own use which puts

- all the font files on in the right place
- ditto for the beamer files
- as well as a small local variant I made for my use.

It is in [this PPA on launchpad](https://launchpad.net/~edd/+archive/ubuntu/misc/+packages).

matze:
#103 should've fixed the problem for our Docker users.

PR #103, opened 2015-06-25
fixed Dockerfile

awalterschulze:
fix for issue https://github.com/matze/mtheme/issues/102


benjamin-weiss:
Some remarks:
- Pygments is no longer needed, as we don't use `minted` in the demo presentation anymore.
- The `build.sh` is still not entirely fixed, as we added references to the demo presentation. So why not use `latexmk` to build the demo presentation? Or couldn't that even be solved by `docker-run: docker-build $(DEMO_PDF)`?
- As I don't really know docker I'm not sure if I'm right with this one. I don't think you build the documentation and add it to the container. Right?

awalterschulze:
Ok we can remove the installation of pygments :)
Ok so we can change xelatex to latetmk?
I mount the whole folder into the container and then simply build inside the container.  Then the output is placed in the folder, since it was mounted.
I don't know if I fully understand what you are trying to say?

benjamin-weiss:
We added a pdf documentation file. So what happens no when you run `make all` the first time is the following:
```bash
# create the sty files
latex mtheme.ins
# create the documentation
latexmk -xelatex -output-directory=.temptex mtheme.dtx
cp .temptex/mtheme.pdf .
# create the demo presentation
latexmk -xelatex -output-directory=.temptex demo.tex
cp .temptex/demo.pdf .
```
You can obviously skip the part with the temporary folder as this is only a way to keep the top directory clean, but as far as I see it you are missing the documentation. And `latexmk`is just a convenient script which takes care of the multiple builds. So you don't have to take care of calling `xelatex` multiple times of your own.

awalterschulze:
This means you could make the build.sh file contain
```
#!/bin/sh
set -xe
latex mtheme.ins
latexmk -xelatex mtheme.dtx
latexmk -xelatex demo.tex
```
or even
```
#!/bin/sh
set -xe
latex mtheme.ins
latexmk -xelatex demo.tex
```

benjamin-weiss:
The first one is right. In the second one you are again missing the documentation.

awalterschulze:
Yes I guess it depends on the usecase.
Do you want to build the whole mtheme or just the user's slides.
But I am happy with either.

awalterschulze:
I have tested both and they work well.

benjamin-weiss:
I don't want to use docker at all. :smirk: But I think it would be useful to have the documentation of the theme.

awalterschulze:
That's my main reason for docker, testable installation documentation.
Its just a coincidence that I can also use it to build.

awalterschulze:
Ok so then yes the docs should also be built.
Are you going to make a pull request or should I?

benjamin-weiss:
Thinking about it again. Why is it necessary to build `demo.pdf` and `mtheme.pdf` in a separate script at all. The best way is probably as I proposed earlier: Just add dependencies to `docker-build`.
```
docker-run: docker-build $(DEMO_PDF) $(MANUAL_PDF)
```
Then if `docker-run` is executed `Make` checks if those files are up to date and if not they are automatically created.

benjamin-weiss:
And of course the sty files.
```
docker-run: docker-build $(DEMO_PDF) $(MANUAL_PDF) $(STY)
```

awalterschulze:
I tested that, but the docker still runs every time and then the docker tells me that the files are up to date.

benjamin-weiss:
I don't know if I have understood you. So it works? What happens if you remove `demo.pdf` `mtheme.pdf` and all `sty` files are they recreated?

awalterschulze:
Ok wait wait, I am a bit tired, I mis read your comment.
```
Thinking about it again. Why is it necessary to build demo.pdf and mtheme.pdf in a separate script at all. The best way is probably as I proposed earlier: Just add dependencies to docker-build.

docker-run: docker-build $(DEMO_PDF) $(MANUAL_PDF)
Then if docker-run is executed Make checks if those files are up to date and if not they are automatically created.
```
The idea is that you should be able to build without having the fonts or even latex installed on your own computer.  These dependencies should only be installed inside the docker.

benjamin-weiss:
Ah. Ok, now I get the point of the `build.sh`. Took me a while. Then another thought how we could get rid of the `build.sh`. Wouldn't it be possible to run `make all` inside the docker?

awalterschulze:
Yes I was just lazy :)
I just need to install make inside the docker.

awalterschulze:
You caught me.

awalterschulze:
It just takes forever to install texlive-full, so when I modify the apt-get line, it takes quite a while to test.

benjamin-weiss:
:grin:

Well then I think this is probably the best way to do it, even though it will take a while to test. Agreed?

> Are you going to make a pull request or should I?
You should do it.

awalterschulze:
Its running :)

awalterschulze:
Ok I'll look at this again tomorrow or monday.

awalterschulze:
I now get the following error:
```
* Redefining document command \oldstylenums with arg. spec. 'm' on line 107.
*************************************************
) (/usr/local/texlive/2012/texmf-dist/tex/latex/fontspec/fontspec-xetex.sty
(/usr/local/texlive/2012/texmf-dist/tex/latex/base/fontenc.sty
(/usr/local/texlive/2012/texmf-dist/tex/latex/euenc/eu1enc.def)
(/usr/local/texlive/2012/texmf-dist/tex/latex/euenc/eu1lmr.fd))
(/usr/local/texlive/2012/texmf-dist/tex/xelatex/xunicode/xunicode.sty
(/usr/local/texlive/2012/texmf-dist/tex/latex/tipa/t3enc.def
(/usr/local/texlive/2012/texmf-dist/tex/latex/euenc/eu1lmss.fd)))
(/usr/local/texlive/2012/texmf-dist/tex/latex/fontspec/fontspec.cfg)))kpathsea: Invalid fontname `Fira Sans Light', contains ' '


!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!
! fontspec error: "font-not-found"
!
! The font "Fira Sans Light" cannot be found.
!
! See the fontspec documentation for further information.
!
! For immediate help type H <return>.
!...............................................

l.26 ...    BoldFont={Fira Sans}]{Fira Sans Light}
```

awalterschulze:
Given this commit
https://github.com/awalterschulze/mtheme/commit/bbb91092762a0a039755c7473eb68b9a2765a604

benjamin-weiss:
Drop the `$(DEMO_PDF) $(MANUAL_PDF) $(STY)`. This was a proposal before I understood what the `build.sh` is actually for.

awalterschulze:
Perfect :)

awalterschulze:
Thank you pushing me, this is much better :)

benjamin-weiss:
Does this also work?
```makefile
docker-run: docker-build
	docker run --rm=true --name $(DOCKER_CONTAINER) -i -t -v `pwd`:/data $(DOCKER_IMAGE) /data/make
```
Then we could delete the `build.sh` file.

awalterschulze:
Thats true
just make instead of /data/make

PR #106, opened 2015-06-27
Removed Pygments dependency and run make all in docker to better test the installation

awalterschulze:
Discussion over here
https://github.com/matze/mtheme/pull/103#event-340858278


awalterschulze:
A better one is coming

Issue #99, opened 2015-06-23
Fix vertical alignment of frame title in all cases


benjamin-weiss:
As the following animation shows the vertical alignment is quite decent for our default small caps lowercase. But when using any other formatting the alignment is way off.

![screenshot_title](https://cloud.githubusercontent.com/assets/9158719/8302951/9df99588-199a-11e5-96b1-b053b89dc974.gif)

tmgriffiths:
See Fix #115

rchurchley:
Great work, @tmgriffiths!

PR #113, opened 2015-07-10
Various issue fixes

rchurchley:
This PR:
- Adds options to put the progress bar at the very top or very bottom of each slide;
- Reverts `itemize` indentation to the beamer defaults; and
- Provides a better solution than the `voffset` hack to vertically center text in a frame.


matze:
If you could use the "Fix #xxx: change foobar" in the future, people without access to GitHub can see immediately what the commit addresses. But I'll let this slip through this time :smile:

matze:
The `\plain` frame content is not centered horizontally anymore. If this is intentionally, we should update the demo to center it when calling `\plain`.

rchurchley:
> If you could use the "Fix #xxx: change foobar" in the future, people without access to GitHub can see immediately what the commit addresses. But I'll let this slip through this time.

Noted. Will do!

> The \plain frame content is not centered horizontally anymore.

I don't think that should be the case... do you have an MWE? I don't seem to be able to reproduce. On my machine,

```latex
\documentclass{beamer}
\usetheme{m}
\begin{document}
\plain{A plain slide}
\end{document}
```
produces the following (expected) output as of bf8d202.
screen shot 2015-07-10 at 9 19 03 am

benjamin-weiss:
I have seen you used `addtobeamertheme`, I cannot test this right now as I'm on the road, but I suppose this breaks mid document changes to these options?

Also you changed the `dp` `ht` structure of the footer with an `sep=3ex`. I like that you do not use centimeters any more and `sep` feels definitely more natural there, but we should probably think about adjusting these margins in the frame title box in a similar way. Maybe adding a `\strut` there and also using `sep` would work.

rchurchley:
> I have seen you used `addtobeamertheme`, I cannot test this right now as I'm on the road, but I suppose this breaks mid document changes to these options?

It actually does not! It turns out that resetting a Beamer template also resets the effects of `addtobeamertheme`. The option can be changed mid-theme without any problem.

> Also you changed the `dp` `ht` structure of the footer with an `sep=3ex`. I like that you do not use centimeters any more and `sep` feels definitely more natural there, but we should probably think about adjusting these margins in the frame title box in a similar way. Maybe adding a `\strut` there and also using `sep` would work. 

Changing the structure of the other Beamer color boxes is a good idea that perhaps we should handle in another issue/pull request.

benjamin-weiss:
> It actually does not! It turns out that resetting a Beamer template also resets the effects of addtobeamertheme. The option can be changed mid-theme without any problem.

Nice. 

> Changing the structure of the other Beamer color boxes is a good idea that perhaps we should handle in another issue. 

Maybe in #99 ?

matze:
> I don't think that should be the case... do you have an MWE? I don't seem to be able to reproduce. On my machine, [...] produces the following (expected) output as of bf8d202.

That is weird. I get this:

glitch

rchurchley:
> That is weird. I get this.

I cloned and recompiled on another machine and still have things centred. I did force push a rebase against master, so perhaps that somehow messed something up for you?

What does the definition of `\plain` look like in your copy of `beamerthemem.dtx`? The correct definition should be as follows:

```latex
\newcommand{\plain}[2][]{%
  \begingroup
    \setbeamercolor{background canvas}{
      use=palette primary,
      parent=palette primary
    }
    \begin{frame}[c]{#1}
      \centering
      \usebeamercolor[fg]{palette primary}
      \usebeamerfont{section title}
      \@metropolis@plaintitleformat{#2}
    \end{frame}
  \endgroup
}
```

benjamin-weiss:
I just tested the PR on my machine and I get a properly centered plain frame. And I cannot spot any problem with the new code for the `plain` macro. Seems fine to me.

matze:
I checked out freshly, removed all remnants and triple-checked that I am on bf8d202, to no avail.

I also removed the theme and only define the `\plain` command in the MWE. As far as I can see – for my Beamer version – horizontal centering only works *if* both `\centering` and the `\vfill`s before and after the inserted text are present.

I currently use Beamer 3.24, could you check if you use a newer version? Unfortunately, the [ChangeLog](http://ftp.fernuni-hagen.de/ftp-dir/pub/mirrors/www.ctan.org/macros/latex/contrib/beamer/ChangeLog) does not indiciate any change in that direction.

rchurchley:
> I currently use Beamer 3.24, could you check if you use a newer version?

I'm using Beamer 3.36, so perhaps there was a bug in previous versions of Beamer that is causing this? I'm not sure what it could be, though. Perhaps your version of Beamer puts another alignment command before the first line or after the last line? If so, it shouldn't be necessary to have `\vfill`s before and after the inserted text, but maybe we need to have something printed in vertical mode before and after the inserted text.

1. What happens if you replace the `\vfill`s with `\par`s?
2. What happens if you use the `center` environment instead of `\centering`?

```latex
\newcommand{\plain}[2][]{%
  \begingroup
    \setbeamercolor{background canvas}{
      use=palette primary,
      parent=palette primary
    }
    \begin{frame}[c]{#1}
      \begin{center}
        \usebeamercolor[fg]{palette primary}
        \usebeamerfont{section title}
        \@metropolis@plaintitleformat{#2}
      \end{center}
    \end{frame}
  \endgroup
}
```

matze:
Both, ending with `\par` and the `center` environment work (now thinking about that I actually had seen this issue with a paper before too). Why this is not a problem for later Beamer versions, we don't know but it would be good to implement either of those solutions (I'd prefer the environment) since 3.24 is the packaged version for Ubuntu 14.04 LTS.

rchurchley:
Bizarre. I also have no idea where the problem arose or when they fixed it, but since it's an easy problem to solve on our end we ought to do so. I agree that the `center` environment is preferable, so I've implemented that fix now.

matze:
Thanks!

PR #115, opened 2015-07-16
Frametitle Fixes

### LINKED ISSUE #99: Fix vertical alignment of frame title in all cases

tmgriffiths:
Hi there,

So this PR should fix #99: 'Vertical alignment of frame title in all cases', also mentioned in a coment on fix #113.

I changed the definition of the `beamercolorbox` to use `sep` rather than `ht, dp, and left/rightskip`. This has the added benefit of making multiline headings possible (objective uglyness and readability aside).

I added a strut and custom length for the height of that strut so that the colorbox always has a minimum height. The height of which is the height of the tallest alphabetic character. The default small caps spacing should thus be unchanged; 1.5ex over and under the text and 1ex of text for a total of 4.5ex. When the `frametitleformat=uppercase/regular` option is given the strut is the height of the tallest uppercase character, so the title should still be a symetrical 1.5ex below and above plus the text in the middle.

Note: 
 - I am using a differnet typeface (Tex Gyre Heros), I was testing more than one thing at time. 
 - The titles also have an added 50/1000 em letterspacing from the microtype package for readability. This is something I'm also going to suggets on readability grounds, but i'll do it in a seperate pull request. The code for that is not in this fork, just the example images.

Old code:
```latex  
 \begin{beamercolorbox}[%
       wd=\paperwidth,
       leftskip=0.3cm,
       rightskip=0.3cm,
       ht=2.5ex,
       dp=1.5ex
      ]{frametitle}
    \insertframetitle
    \end{beamercolorbox}
```
With currrent definition (`ht=2.5ex, dp=1.5ex, leftskip=0.3cm, rightskip=0.3cm`)

oneline_no-sep
twoline_no-sep

---

New code:
```latex
   \begin{beamercolorbox}[%
       wd=\paperwidth,
       sep=1.25ex,
      ]{frametitle}
    \insertframetitle\rule{0pt}{\@metropolis@frametitlestrut}
    \end{beamercolorbox}
```
With sep parameter (`sep=1.5ex`), `frametitleformat=lowercase` option

oneline_with-sep
twoline_with-sep

---
With the `frametitleformat=uppercase` option

uppercase_oneline_with-sep
uppercase_twoline_with-sep

---
With the `frametitleformat=regular` option

regularcase_oneline_with-sep
regularcase_twoline_with-sep


matze:
Looks and works perfectly fine for me. Do you see any issues @benjamin-weiss and @rchurchley?

rchurchley:
This is fantastic and exactly the sort of thing we needed to solve #99!

If I could make one suggestion: I would prefer it if the frametitleformat options redefined a command instead of set a length. That is, define `\@metropolis@frametitlestrut` as the strut instead of the height of the strut, and change them with the following.

```latex
 \renewcommand{\@metropolis@frametitlestrut}{%
    \rule{5pt}{\heightof{abcdefghijklmnopqrstuvwxyz}}
}
```
This is just for readability purposes. The code in this pull request works, but it wasn't obvious to my why it should work until I realized that `\setlength` does not expand its arguments before storing the length.

tmgriffiths:
Thanks @rchurchley! I've changed from a \setlength` to a `\renewcommand`.

matze:
Could you rebase against master (and fix occuring conflicts)? As of now, I am unable to merge your PR.

rchurchley:
Thanks for the fast update! Two more notes:

- It looks like a `\myletterspacing` snuck into your last commit from your microtype experiments — could you remove it?  
- Should the strut come before the `\insertframetitle` instead of after? I can think of (admittedly bizarre) corner cases where this might make a difference — an extremely long frame title only containing short characters might have its first line too short? It probably doesn't matter much, but I thought I'd bring it up just in case.

benjamin-weiss:
The height calculation is not perfect yet. E.g. if you use the following title frame settings the height of the frame title box varies depending on the depth of the content.
```TeX
\usetheme[everytitleformat=regular,sectiontitleformat=regular]{m}
\setbeamerfont{frametitle}{family=\Book, size=\large, shape=\upshape}
```
I don't have a proper solution for this yet. But apart from that, this is exactly what I had in mind. Great work!

PS: Letter spacing can also be done with font features of `fontspec`. This is maybe worth a try.
```TeX
\setmainfont[SmallCapsFeatures={LetterSpace=6}]{Fira Sans Light}
```

rchurchley:
>  if you use the following title frame settings the height of the frame title box varies depending on the depth of the content

@benjamin-weiss, can you give a full MWE for this? The height of the frame title box does *expand* to fit the user's frame title — for example, if it extends over multiple lines, or if it contains large non-alphabetic symbols in `lowercase` mode — but I have not been able to produce an example where a frametitle has a shorter height than usual.

fonttest

As far as I can tell, this PR should be mergeable after four quick cleanup items:
- rebase onto the current master,
- move `\@metropolis@frametitlestrut` before `\insertframetitle` (?)
- remove `\myletterspacing` from the `uppercase` option code
- replace `\newlength{\@metropolis@frametitlestrut}` to `\providecommand{\@metropolis@frametitlestrut}`

If I've overlooked an example that violates the minimum frametitle height, then we should find a fix. Otherwise, if @benjamin-weiss's problem involves the frametitle unexpectedly expanding beyond the minimum height, I think we should merge this PR and open a new issue to discuss that. The letter spacing question should also probably be moved to a new issue as well.

tmgriffiths:
I’ve put a `\@metropolis@frametitlestrut` at the beginning and the end because of what I can see will be an error for multiline titles. Perhaps what @benjamin-weiss was alluding too?

If we were to put the strut at the end of a multiline title, then if the first line was nothing but lower case characters (for some reason…) in a `frametitleformat=regular` situation then the strut would make sure that the minimum height would still be that of a capital letter. But, since this is a multiline title that only happens for the last line. 

If we thing of the opposite situation; the strut is at the beginning but our final line is nothing but lowercase. Then the strut again makes sure that the minimum height is that of a capital letter. But, this is only applied to the first line now. To rectify this I’ve put a strut at both the beginning and the end. Now, this won’t help If the presenter has a multiline title that is three or more lines long (god forbid) but in that case the standard leading and `\baselinskip` should take care of it. Also, the central line(s) shouldn’t effect the padding around the title anyway.

This should address the items you mentioned @rchurchley.

benjamin-weiss:
> I’ve put a \@metropolis@frametitlestrut at the beginning and the end because of what I can see will be an error for multiline titles. Perhaps what @benjamin-weiss was alluding too?

Yes, that's what I meant.

> can you give a full MWE for this? 

If I add the frame title format I posted before to the demo I get different sized frame title boxes.

height

The problem is, that the custom strut is not properly handling the depth of the font. It doesn't really matter for small caps or all upper case, but for a regular font shape and upper and lowercase letters this is a problem.

height2

rchurchley:
> the strut is at the beginning but our final line is nothing but lowercase. Then the strut again makes sure that the minimum height is that of a capital letter.

I would have guessed that `\baselineskip` would also do the job here, without need for a final strut. But putting the strut before the title satisfies my nitpicking!

> If I add the frame title format I posted before to the demo I get different sized frame title boxes.

I'm not sure why my previous tests didn't reproduce this issue, but I can see the problem now. I believe the solution is to use `\vphantom` instead of `\rule`, and make sure that the code for `regular` also contains characters with descenders:

```latex
regular/.code={%
  \renewcommand{\@metropolis@frametitleformat}{}%
  \renewcommand{\@metropolis@frametitlestrut}{%
    \vphantom{ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz}
  }
},
```
Does that fix the problem, @benjamin-weiss?

benjamin-weiss:
This is definitely a better solution, as it fixes the different sized frame title boxes. But the solution is still not perfect imho.

height3
I think either `a_1` or `a_2` should be equal to `b` and the same goes for `c`. What do you think?

rchurchley:
> I think either a_1 or a_2 should be equal to b and the same goes for c. What do you think?

After some testing, I believe the discrepancies between these lengths are due to the natural behaviour of `sep` and not to @tmgriffiths' implementation. I still think `sep`'s behaviour is preferable to the current `master` so I'd vote to merge this the changes in this pull request and move discussion of the spacing to a new issue.

benjamin-weiss:
>After some testing, I believe the discrepancies between these lengths are due to the natural behaviour of `sep` and not to @tmgriffiths' implementation. 

It is both. `a_1` and `c` we can easily set to the same length by properly ending the lines with `%`. 

```TeX
\newlength{\@metropolis@frametitlestrut}
\defbeamertemplate{frametitle}{plain}{%
  \nointerlineskip%
  \begin{beamercolorbox}[%
      wd=\paperwidth,
      sep=1.5ex,
    ]{frametitle}%
  \@metropolis@frametitlestrut\insertframetitle\@metropolis@frametitlestrut%
  \end{beamercolorbox}%
}
```
Currently we just add an additional space char. But setting `b` equal to `a_1` is due to the behavior of sep and not so easy to resolve.

rchurchley:
Ah yes. So this PR is ready to merge once @tmgriffiths adds the missing `%`?

benjamin-weiss:
Agreed.

matze:
Thanks for the effort! :+1:

rchurchley:
Thanks for your hard work and patience, @tmgriffiths!

PR #122, opened 2015-08-18
Correct plaintitleformat in documentation

ptheywood:
The main theme `plaintitleformat` option was incorrectly documented as `plainformat`.


matze:
Thanks!

PR #132, opened 2015-09-22
Reorganize repository into folders

rchurchley:
I had the opportunity to use this theme with some novice LaTeX users over the summer. They ran into some confusion over:

1. How to install the theme without `make` on their system,
2. Which files they needed to put in their shared Dropbox folder to use the theme without installing it on all their computers, and
3. What prerequisites were actually needed to use the theme (rather than to compile the demo).

This PR attempts to address these problems — and, admittedly, to satisfy my own organizational desires — by arranging the repository's files into subfolders. I have also adding more complete installation instructions to the README and manual.

- The demo slides are added to a `demo` subfolder, reducing the probability that the user will compile it to test their system's compatibility with the Metropolis theme. (If they do, they may come to the erroneous conclusion that they need to install `pgfplots` and `ccicons` to use Metropolis.)
- The source `.ins` and `.dtx` files are added to a `source` folder, the docker files to a `docker` folder, and the manual to a `doc` folder. However, `make sty` still compiles the `.sty` files to the top-level directory. This increases the probability that the user will only move the `.sty` files to their presentation's working directory.
- The README and manual now have updated and more detailed installation instructions for new users.

What do you think?


rchurchley:
Note: I am not familiar with Docker, so the Docker-related rules in the Makefile might need to be updated to work with the new directory structure.

ptheywood:
In the updated readme you seem to have some LaTeX left over in a few places `\themename{}` & `\themename` but otherwise reorganising the repository seems like a good idea to me.

matze:
> What do you think?

Useful changes! If you fix the issues pointed out by @ptheywood, I'll merge the changes.

rchurchley:
> In the updated readme you seem to have some LaTeX left over in a few places \themename{} & \themename

Oops! Now fixed.

> One more thing: the ctan target is broken right now.

Thanks for catching that. I've fixed it now (along with a more subtle TEXINPUTS problem that led `make demo` to omit the bibliography).

While I was at it, I adjusted the behaviour of `make ctan`. Previously, it named the zip folder based on the date given in `beamerthemem.dtx`. With this PR, `make ctan` now *updates* the date in the package files and names the zip file accordingly with the current date. I hope this behaviour is all right!

I believe all targets should build properly now, with the possible exception of the `docker` rules which I have not changed or tested.

matze:
Alright, thanks for fixing these issues!

> With this PR, make ctan now updates the date in the package files and names the zip file accordingly with the current date. I hope this behaviour is all right!

I think it is. In any case, for a proper release we still have to update the version number manually but that cannot be avoided.

Issue #130, opened 2015-09-12
\metroset{background=dark} does not invert all text colors on title slide


maxheld83:
Oddly, `\metroset{background=dark}` does not affect all parts of the title slide, but only the title and subtitle.
The font color of the rest remains unchanged.

The behavior is different when using `background=dark` as an option; in that case, *all* colors are inverted appropriately on the title slide.

maxheld83:
oddly, setting `\setbeamercolor{normal text}{fg = white, bg = white}` inside that group also does not change the color of date, institute etc.

maxheld83:
I should add: I've set `\texorpdfstring{}...` inside `author{}`, such as:

```
\author[Held \& Zabala]{
  \texorpdfstring{
    \begin{columns}
      \column{.45\linewidth}
      %\centering
      Maximilian Held\\
      \href{mailto:info@maxheld.de}{info@emaxheld.de}\\
      \href{http://www.maxheld.de}{www.maxheld.de}\\
      \column{.45\linewidth}
      %\centering
      Aiora Zabala\\
      \href{mailto:aiora.zabala@gmail.com}{aiora.zabala@gmail.com}\\
      \href{http://people.ds.cam.ac.uk/az296/}{people.ds.cam.ac.uk/az296/}\\
    \end{columns}
  }{
    Maximilian Held \inst{1}
    \and
    Aiora Zabala \inst{2}
  }
}
```

adding this line fixes things:

```
 \usebeamercolor[white]{author}
```

Notably ` \usebeamercolor[fg]{author}` *does not work in this place*

rchurchley:
Thanks for the report! I think I've isolated the issue and submitted a pull request to fix it.

treuherz:
Still getting this problem with a normal slide. Colours invert fine across the presentation when it's set in the preamble, but when using \metroset{background=dark} before a specific slide I want to look different, the title renders fine but all body text is invisible.

benjamin-weiss:
This seems to be a bug in beamer itself. I encountered this problem on another occasion. The workaround mentioned [on StackExchange](http://tex.stackexchange.com/questions/288408/beamer-show-notes-on-second-screen-with-xelatex-and-atbeginsection) seems to be also working here.

```TeX
\documentclass{beamer}
\usetheme{metropolis}

\makeatletter
\def\beamer@framenotesbegin{% at beginning of slide
  \usebeamercolor[fg]{normal text}
  \gdef\beamer@noteitems{}%
[...]

PR #133, opened 2015-09-28
Fix title page templates

rchurchley:
As reported by @maxheld83 in #130, the `author`, `date`, and `institute` on the title page do not update their colours when `normal text` is changed with `\metroset`. This PR fixes this by adding the appropriate calls to `use=normal text` in the colour theme.

I've also bundled into this PR a reorganization of the title page code, which helped track down this problem and will also be useful for future improvements.


matze:
Thanks for taking care of #130.

PR #142, opened 2015-10-19
Allow installation in any directories

luc-j-bourhis:
I would like to create packages to install mtheme on various Linux distros using the native package manager (apt-get, yum, paceman, etc). This makes it very easy to install the Fira fonts as dependencies. In order to achieve that, I need to be able to install the files in the staging area. Hence the patch to Makefile.


matze:
Looks good to me :+1:

PR #145, opened 2015-10-24
Fixes #85 – Font theme overhaul

### LINKED ISSUE #85: Use the `fira` bundle
Adding this issue to consolidate discussion about the [`fira`](http://www.ctan.org/pkg/fira) bundle.

In #75, @ChipmunkMath made us aware of these packages:

> [`fira`](http://www.ctan.org/pkg/fira) // Fira Sans Fonts: Provides Fira Sans and Fira Mono fonts, and works outside of xelatex/lualatex. I was playing around with it and typesetting just in plain pdflatex. Seems to be in pretty good shape, although it's important to make sure you're updated to the most recent version (2015-05-23) which fixes some major gaps.

The included packages — `FiraSans.sty` and `FiraMono.sty` purport to provide the Fira fonts for pdflatex, xelatex, lualatex, etc. If it works, it would remove the need for users to find and install the Fira fonts, which is the biggest barrier for a prospective user to adopt this theme. It would allow us to remove all the platform-specific instructions for installing the Fira fonts or manually changing the theme if Fira has an unexpected name. And as an added bonus, it would allow full compatibility with `pdflatex` for the conservative users who don't want to use `xelatex` for whatever reason.

As @ChipmunkMath points out, though

> There is a potential issue in that fira is extremely recent, so unless a user has been doing a good job keeping their TeX install up to date, they might not have the package. Similarly, even if they have the package, it's important for them to be on the most recent version (extremely recent, less than a month ago), as the older version was missing some important things. But it should be easier to tell a potential new user to just go run TeX Live (or equivalent) and make sure they have the most current version of fira than having them do the current install method.

> As a caveat, though, I'd recommend extensive testing with the fira package before adopting it into the metropolis theme as a required package. When I was playing with it before, it didn't seem to be 100% perfect. It was very close, but there were some minor issues, so we should be careful before choosing to work with it.

This issue is for discussion on the `fira` bundle and whether we should use it instead of `\fontspec` and the user's font domain for Fira Sans. @ChipmunkMath , do you have any details on the minor issues you came across with `fira`?

benjamin-weiss:
I finally started working on the font theme. The most important change in this PR is that all the `\setbeamerfont{•}{•}` commands are solely defined via standard LaTeX commands. The downside of this is, that now only 2 font weights can be used. But the `\Medium` font weight hasn't been used anyway and the `\ExtraLight` font weight was already planed to be removed (#95). But it is now very easy to change the font. A simple 
```TeX
\setsansfont[•]{•}
\setmonofont[•]{•}
```
is sufficient to change the fonts.

I wasn't able to get the Fira package to work properly in its *otf* mode. Only in the *type1* mode I found a way to use the light font weight as standard and the regular font weight as bold.
```TeX
\RequirePackage[type1]{FiraSans}
\def\seriesdefault{l}
\def\bfseries@sf{m}
\RequirePackage[type1]{FiraMono}
```
This works very well with both pdfLaTeX and Xe-/LuaTeX. But these definitions are a problem if one tries to load an other font via `fontspec` after the theme is loaded. Then some commands have to be executed to get back to the default behavior.
```TeX
\def\seriesdefault{\mddefault}
\def\bfseries@sf{bx}
\def\bfseries@tt{bx}
```
Any ideas how to handle that?


benjamin-weiss:
Ok, I am getting really annoyed by the `fira` package. I just realized, that some characters are interchanged (at least in case of the type1 mode).

demo-2 verschoben

So here is another proposal. We just keep it as it was. If the theme is compiled with the XeLaTeX or LuaTeX the fira font will get loaded in every other case not. And to make the font available on every system without having the user install it, we just add the font files to the package. As far as I can see the license used for the fira font should allow this.

rchurchley:
I like the way you've moved the conditional fontspec inclusion into the font theme itself; this looks exactly like the overhaul we need. If the `fira` package doesn't work the way we need it to, I think it's all right to abandon it.

Two possible alternatives to adding the fira fonts to the package:
- [load the Fira fonts only if they exist on the user's system](http://tex.stackexchange.com/questions/53443/how-to-check-if-a-font-exists-in-xelatex)
- use a [git submodule](https://git-scm.com/docs/git-submodule) to point to the [official Fira repo](https://github.com/mozilla/Fira). (Submodules are often more trouble than they are worth, but this appears to be one of the few use cases where they actually are the right answer.)

I don't have a particular opinion on whether we should include the Fira fonts directly or use one of the above alternatives, but I thought I should bring them up for discussion. However, if we do redistribute the font files I believe we are required to also add [their license file](https://github.com/mozilla/Fira/blob/master/LICENSE) to the `fira` directory.

matze:
We should also keep an eye on integration with Linux distributions. They will usually package Fira themselves and create proper dependencies between the theme and the font. Including the font ourselves might cause some unwanted conflicts. But I am all for an easier use of the theme + Fira for all users.

benjamin-weiss:
I experimented a while back with conditional font loading in XeLaTeX and also came across the SE question you linked. At the time I tested it, it resulted in a very long compile time if the font wasn't found. But I just tested my test tex file again and this doesn't seem to be the case anymore. Here is my tex file.

[fonttest.tex](https://gist.github.com/benjamin-weiss/dcf6a5ee82adbfb514f2)

I also thought about pointing to the official fira repository. But I never used that and was to lazy to look into it. :smirk: But I think this makes a lot of sense.

Concerning the license you are probably right. At least it doesn't hurt to do it.

benjamin-weiss:
I dont't see how my current solution could lead to unwanted conflicts, as it doesn't matter whether the font is installed or not. It just eliminates the dependency.

benjamin-weiss:
So is looked into git submodules. Although it works nice, there are two downsides.

* The first one is not a big deal, but could lead to confusion for people who are not that familiar with git. The repository would have to be cloned with the `--recursive` option: `git clone --recursive https://github.com/matze/mtheme`
* The second is annoying: The Fira font repository is a massive 185.1 MB. Cloning nearly 200 MB for just  1.8 MB of otf files seems a bit over the top.

So in my opinion the current solution is the best one. What do you guys think?

matze:
> So in my opinion the current solution is the best one. What do you guys think?

Yes, it's better than pulling the whole repository in. However, I still see the issue with regards to packaging. See, for example the [package](https://aur.archlinux.org/packages/texlive-mtheme/) for Arch Linux. They use the `install` target to install the theme. But with your current change, they will end up with two competing Fira installations. The system Fira (as per the specified dependency) and the one we install into the TeX directory. I know that we live in an age of virtually unlimited disk space but redundancy is an issue concerning expectation and updates. How to fix this: I don't know :-(

benjamin-weiss:
> However, I still see the issue with regards to packaging.

* First a package like the Arch Linux one should obviously remove the Fira font dependencies. 
* Second once the CTAN release is finished all Linux packages are imho anyway obsolete.
* Third try not to see them as competing installations. :smirk: 

But there is a solution that might satisfy you. We could add a Makefile option, that creates a different `beamerfonttheme.sty` file, which loads the Fira fonts installed in the system, and does not install the Fira fonts. But to me this feels a little bit overcomplicated.

rchurchley:
> Second once the CTAN release is finished...

Actually, come to think of it, when we start using CTAN for distribution then including the fonts will not be helpful. If `tlmgr` does not install them automatically, then they won't provide any benefit over linking to [the Fira website](http://www.carrois.com/fira-4-1/) in the documentation. But if `tlmgr` does install them automatically, then as @matze points out we start having to worrying about the user's environment, conflicts with existing copies of the Fira fonts (possibly under different names), and so forth. In addition, redistributing the fonts means somebody would have to keep an eye on the Fira project to stay on top of any updates they release.

I'm starting to think that including the fonts directly in the repo may be more trouble than they're worth.

rchurchley:
> I experimented a while back with conditional font loading in XeLaTeX and also came across the SE question you linked. At the time I tested it, it resulted in a very long compile time if the font wasn't found. But I just tested my test tex file again and this doesn't seem to be the case anymore.

I think it would be great to conditionally include the Fira fonts then. Slow compilation gives a better first impression than crashing due to missing fonts, and if the compilation is still fast then all the better! Conditional font loading would also enable us to support differently-named Fira fonts instead of asking users to modify the font theme (as we currently have to).

benjamin-weiss:
I don't see where there should be a problem if the `tlmgr` installs the font files with the theme (into the same folder as the `sty` files) as only the theme will access those files. It is the same with the `fira` package there are also no conflicts.

rchurchley:
> I don't see why there should be a problem if the `tlmgr` installs the font files with the theme (into the same folder as the `sty` files) as only the theme will access those files.

Suppose I already have Fira installed. If I install Metropolis and use it in a Beamer document, which copy will take precedence? This won't make a difference if the files are the same, but could cause headaches as new versions of Fira are launched with updated glyphs or an expanded character set. 

If the `texmf` copy takes precedence, a XeLaTeX user might update the font through their OS's package manager but never realize they also need to update Metropolis (assuming we've kept up-to-date on the Fira version). On the other hand, if the user's copy takes precedence normally, the font file used will confusingly depend on the user's OS, since the `texmf` version of Fira Sans will still beat out a user-installed Fira Sans OT.

benjamin-weiss:
Ok, but I wouldn't call that a conflict. It is just something which could lead to confusion for the user.

In the current implementation the `texmf` copy takes precedence. And if somebody wants to use the systems Fira font, he would just have to add two simple lines.
```TeX
\setsansfont[BoldFont={Fira Sans}]{Fira Sans Light}
\setmonofont{Fira Mono}
```
To make it even easier we could add a theme option for that.

And a word with respect to the first impression. There I'm also convinced, that the current implementation is quite good, because the user doesn't have to install the font manually. If the metropolis package is installed and the presentation is compiled with XeLaTeX or LuaTeX he can immediately use the Fira font.

benjamin-weiss:
So what are we gonna do? I think the `fira` package is off the table, right? So we are left with the proposals *installation of the otf files with the package* and *conditional font loading*. I updated the conditional font loading [gist](https://gist.github.com/benjamin-weiss/dcf6a5ee82adbfb514f2) with another macro to check multiple fonts at once. So a conditional font loading scheme would look something like this (untested).

```TeX
\ifboolexpr{bool {xetex} or bool {luatex}}{
    \RequirePackage[no-math]{fontspec}
    \defaultfontfeatures{Mapping=tex-text}
    \iffontsexist{Fira Sans Light, Fira Sans Light Italic, Fira Sans, Fira Sans Italic}{%
        \setsansfont[BoldFont={Fira Sans}]{Fira Sans Light}
    }{%
        \iffontsexist{Fira Sans Light OT, Fira Sans Light Italic OT, Fira Sans OT, Fira Sans Italic OT}{%
            \setsansfont[BoldFont={Fira Sans OT}]{Fira Sans Light OT}
        }{%
            \PackageWarning{beamerthemem}{%
                 Could not find Fira Sans fonts.
            }
        }
    }
    \iffontsexist{Fira Mono, Fira Mono Bold}{%
        \setmonofont{Fira Mono}
    }{%
        \iffontsexist{Fira Mono OT, Fira Mono Bold OT}{%
            \setmonofont{Fira Mono OT}
        }{%
            \PackageWarning{beamerthemem}{%
                 Could not find Fira Mono fonts.
            }
        }
    }
    \AtBeginEnvironment{tabular}{%
        \addfontfeature{Numbers={Monospaced}}%
    }
}{%
    \PackageWarning{beamerthemem}{%
         You need to compile with XeLaTeX or LuaLaTeX to use the Fira fonts.
    }
}
```
Actually I'm fine with both solutions. I don't really care, because I usually load a different font anyway.

rchurchley:
To be honest, I don't have a strong opinion on this either; I'd be fine with either one. @matze?

matze:
(Sorry, have been absent during the weekend)

Conditional font loading looks like an acceptable solution to me, so please go on.

benjamin-weiss:
Alright, so I will finish this PR with a conditional font loading scheme. I will drop you a line when I'm done.

benjamin-weiss:
Done.

matze:
Could you squash some of the commits (i.e. those that add and remove the fonts)? This will avoid blowing up the repo size unnecessarily. Other than that, good work I'll merge :+1:

benjamin-weiss:
Yet another possibility to train my git skills. Yay! ;)

Squashed it to one commit.

matze:
Very good, thanks :-)

msbauer:
I have been trying to use the ``fira`` package as well to compile with pdflatex and found the same annoying interchanges that @benjamin-weiss reported.

However, I found that 
```
\ifboolexpr{bool {xetex} or bool {luatex}}{
...
}{
\RequirePackage[sfdefault,light,lining]{FiraSans}
\RequirePackage[nomap, lining]{FiraMono}
\RequirePackage{mathtools}
\def\bfseries@sf{m}
\RequirePackage[T1]{fontenc}
}
```
solves the problem.

PR #148, opened 2015-11-13
fixes #131 – Wrong frametitle spacing with tailed letters

### LINKED ISSUE #131: Wrong frametitle spacing with tailed letters
First of all thank you for your work, I just started using your theme and love it!

The issue I noticed is that the frametitle box has a different height when using the Fira fonts and when the frametitle contains the letters "Q", "Y", etc. due to their tails: it's just as if the rule printed by the command \@metropolis@frametitlestrut had the wrong height (maybe the fonts are not correctly loaded when defining it?).

Thanks in advance!

Edit: obviously the issue is gone when replacing \@metropolis@frametitlestrut with LaTeX default \strut.

Edit2: this figure shows what's going on with the demo.
metropolis

benjamin-weiss:
`vphantom` implementation proposed by @rchurchley in #115


matze:
I cannot merge as is, due to a merge conflict.

benjamin-weiss:
I resolved the merge conflict.

matze:
Thanks :+1:

Issue #152, opened 2015-11-22
Cannot find fira fonts


bklamer:
The change to conditional font detection has broken metropolis for my installation. I am on ubuntu 14.04 with TexLive 2015 and have fira v4.1 installed. When compiling with lualatex, I receive a warning that the fira fonts cannot be found and they are not used in the resulting pdf. It worked fine with the previous font methods.

benjamin-weiss:
Without checking I just assumed the font handling would work exactly the same for XeLaTeX and LuaTeX. But I can reproduce your problem with LuaTex. So there must be some difference. But as of now I still couldn't figure out what the actual problem is. 

As a workaround until it is fixed you can load the fira font manually after you loaded the theme.
```TeX
\documentclass{beamer}
\usetheme{metropolis}
\setsansfont[BoldFont={Fira Sans}]{Fira Sans Light}
\setmonofont{Fira Mono}
\begin{document}
\begin{frame}{Title}
	Content
\end{frame}
\end{document}
```

benjamin-weiss:
Can you please check if [this patch](https://github.com/benjamin-weiss/mtheme/tree/font-loading-fix) fixes your problem?

bklamer:
Yes, it compiles normally again. thanks

PR #154, opened 2015-11-22
Fix font features for monospace fonts

bklamer:
Ligatures should not be turned on for monospaced fonts (you don't want curly quotes in code fragments and other special features). Also, `Ligatures=TeX` is the same as having `Mapping=tex-text`, but is more compatible with luatex (See sections 9.1 and 11.1 of the fontspec manual).

### DIFF
--- a/source/beamerfontthememetropolis.dtx
+++ b/source/beamerfontthememetropolis.dtx
@@ -42,7 +42,7 @@
 %    \begin{macrocode}
 \ifboolexpr{bool {xetex} or bool {luatex}}{
   \RequirePackage[no-math]{fontspec}
-  \defaultfontfeatures{Mapping=tex-text}
+  \defaultfontfeatures[\rmfamily,\sffamily]{Ligatures=TeX}
 %    \end{macrocode}
 %
 % To simplify the check whether the |Fira| fonts are installed, a set macros is


benjamin-weiss:
Good point. I haven't thought about that line of code yet. So I looked into it a bit. Wouldn't it be more sensible just to remove this line? Because your new line is already the default setting in `fontspec.cfg`. And if somebody really changes his `fontspec.cfg` we probably shouldn't mess with this setting either as it would just lead to confusion.

bklamer:
Your right, turns out it is the default setting. I will close this and create a new pull request to remove the line as suggested.

benjamin-weiss:
If you want you can also incorporate my fix for issue #152 (in `beamerfontthememetropolis.dtx` and `metropolistheme.dtx` → my quick fix only has the changes in `beamerfontthememetropolis.dtx`), then @matze will only have to merge one PR.

And another thing, you should rebuild the documentation and add the updated PDF to your commit to keep it up-to-date.

bklamer:
Sorry, I don't know.

benjamin-weiss:
I had a look at the change history of `fontspec` and it seems like `Ligatures=TeX` is a default setting since version 2.4. So if we want to be backwards compatible to versions earlier than 2.4 we should add
```
\defaultfontfeatures[\rmfamily,\sffamily]{Ligatures=TeX}
```
after all. But other than that ` Ligatures=TeX` and `Mapping=tex-text` should be exactly the same, as the documentation clearly states:

>Using the `tex-text` mapping is also equivalent to writing `Ligatures=TeX`. The use of
the latter syntax is recommended for better compatibility with LuaTEX documents.

And these sentences in the documentation haven't changed since 2010 (Version 2.0).

matze:
Ah, sorry, I was confused by the commit message which I mistakenly took for the actual change. You're both right, that `Ligatures=TeX` is equivalent to mapping. Unfortunately, with fontspec 2.3c I cannot restrict `Ligatures=TeX` to `\rmfamily` and `\sffamily`:
```
(/home/matthias/texmf/tex/latex/metropolis/beamerfontthememetropolis.sty
Package: beamerfontthememetropolis 2015/12/04 Metropolis font theme

! TeX capacity exceeded, sorry [input stack size=5000].
\size@update ->\baselineskip \f@baselineskip 
                                             \relax \baselineskip \f@linespr...
l.84 }
```

@wspr: It seems to be a fontspec problem. This MWE does not compile for

```latex
\documentclass[a4paper]
\usepackage{fontspec}
\defaultfontfeatures[\rmfamily,\sffamily]{Ligatures=TeX}
\begin{document}
``Hey'' --- Test
\end{document}
```

PR #156, opened 2015-11-23
fixes #152 – fixes conditional font loading for LuaTeX

### LINKED ISSUE #152: Cannot find fira fonts

benjamin-weiss:


matze:
Did this change really affect the output of the PDF?

benjamin-weiss:
Sure, as we append the code to the documentation, every piece of code we change changes the output of it.

matze:
Yes, of course. But I meant if this had a real effect on the visual outcome which I doubt. Unfortunately, Git is not very good at version controlling binary data and I really want to avoid growing the repo unnecessarily. As of now, a fresh clone already wastes several MB of bandwidth for some TeX files and two (rather smallish) PDFs. But I'll let that slip :-)

benjamin-weiss:
Ok, now I understand what you mean. Maybe we should just exclude the PDFs from the repository once the CTAN release is finished and just link to the PDFs hosted by CTAN.

PR #160, opened 2015-12-03
fixes #119 – Spacing of blocks in columns, minipages

### LINKED ISSUE #119: Spacing of blocks in columns, minipages etc. is off
While [cleaning up the demo](https://github.com/matze/mtheme/commit/e4764c6642f06cefa32454588d9948037cdff4ab), I noticed an issue with `block` environments within columns or minipages. The spacing of the block content is off and is noticeably bad for blocks with background enabled. The following MWE illustrates the problem

```latex
\documentclass[10pt]{beamer}
\usetheme{m}

\begin{document}
\begin{frame}{Blocks in columns}
  \begin{columns}
    \column{1.0\textwidth}
      \begin{block}{Inside}
        Content.
      \end{block}
  \end{columns}

  \begin{block}{Outside}
    Content.
  \end{block}

  \metroset{block=fill}
  \begin{columns}
    \column{1.0\textwidth}
      \begin{block}{Inside}
        Content.
      \end{block}
  \end{columns}

  \begin{block}{Outside}
    Content.
  \end{block}
\end{frame}
\end{document}
```

blocks

benjamin-weiss:
I decided to reset `\parskip` at the beginning of every `block` environment. So this also fixes the problem for minipages and so on.

### DIFF
--- a/source/beamerinnerthememetropolis.dtx
+++ b/source/beamerinnerthememetropolis.dtx
@@ -393,6 +393,7 @@
 %    \begin{macrocode}
 \newlength{\@metropolis@blockskip}
 \setbeamertemplate{block begin}{%
+  \setlength{\parskip}{\@metropolis@parskip}
   \vspace*{1ex}
   \begin{beamercolorbox}[%
     ht=2.4ex,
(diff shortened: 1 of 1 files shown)


matze:
Ah, what a relief! Thanks :+1:

PR #161, opened 2015-12-04
introduces new title format commands

benjamin-weiss:
I set all the default title formats to `regular` now. I hope that's ok.


matze:
Thanks a lot for that comprehensive rework!

PR #166, opened 2015-12-21
Added option to add a footer to the slides

davidcortesortuno:
I added an option to write a footer to the bottom left part of the frames (it is empty by default). I did it because it is sometimes useful to show the name of a conference, a logo or something similar

### DIFF
--- a/demo/demo.tex
+++ b/demo/demo.tex
@@ -16,12 +16,9 @@
 \date{\today}
 \author{Matthias Vogelgesang}
 \institute{Center for modern beamer themes}
-% \titlegraphic{\hfill\includegraphics[height=1.5cm]{logo/logo}}
+\titlegraphic{\hfill\includegraphics[height=1.5cm]{logo.pdf}}
 
 \begin{document}
-
-% Set a custom footer text for all the slides
-\metroset{footer={custom=My footer}}
 
 \maketitle
 
@@ -282,6 +279,14 @@ or show \textbf{bold} results.\end{verbatim}
     Veni, Vidi, Vici
   \end{quote}
 \end{frame}
+
+{%
+\setbeamertemplate{frame footer}{My custom footer}
+\begin{frame}[fragile]{Frame footer}
+    \themename defines a custom beamer template to add a text to the footer. It can be set via
+    \begin{verbatim}\setbeamertemplate{frame footer}{My custom footer}\end{verbatim}
+\end{frame}
+}
 
 \begin{frame}{References}
   Some references to showcase [allowframebreaks] \cite{knuth92,ConcreteMath,Simpson,Er01,greenwade93}
--- a/demo/demo.tex
+++ b/demo/demo.tex
@@ -20,6 +20,9 @@
 
 \begin{document}
 
+% Set a custom footer text for all the slides
+\metroset{footer={custom=My footer}}
+
 \maketitle
 
 \begin{frame}{Table of contents}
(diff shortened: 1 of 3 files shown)


benjamin-weiss:
I think it is a really good idea to implement something like that, so thank you for your work. But your implementation is imho a bit overcomplicated. I think it is sufficient to just define a new beamer template (like you did) and let the user set the template directly.

Why did you add the `\makebox[0pt][l]{}`?

As for the demo presentation: I would prefer it if there where only one frame with a custom footer and a short explanation how to set it.

davidcortesortuno:
> Why did you add the \makebox[0pt][l]{}?

I just wanted to be sure the footer to be to the left, but it might be not necessary. I can't remember exactly why I did it, maybe because I put the `usebeamertemplate` after the `hfill` command.

> As for the demo presentation: I would prefer it if there where only one frame with a custom footer and a short explanation how to set it

I agree, should I correct that? Although my idea was to just add a comment. 

I also added a comment in the documentation about if it was a good idea to use `footer={custom=My text}`, it seems to work, but I'm not an expert on LaTeX so,  may there be a better solution?

benjamin-weiss:
>I just wanted to be sure the footer to be to the left, but it might be not necessary. I can't remember exactly why I did it, maybe because I put the `usebeamertemplate` after the `hfill` command.

I removed it and it seems to work just fine without it. And I don't see a reason why it shouldn't.
> I also added a comment in the documentation about if it was a good idea to use `footer={custom=My text}`, it seems to work, but I'm not an expert on LaTeX so, may there be a better solution?

As I already tried to explain: I don't think it is necessary to use the theme options at all in this case. I don't see a benefit there. Whether the user has to write:
```TeX
\metroset{footer={custom={My Footer}}}
```
or
```TeX
\setbeamertemplate{frame footer}{My Footer}
```
doesn't really matter. But the latter is imho more "beamer-like" and it drastically simplifies the implementation. @matze What's your take on that?
> I agree, should I correct that? Although my idea was to just add a comment.

As the idea of the theme is to "minimize noise and maximize space for content" the demo presentation should reflect that and not show a footer on every slide. So, yes, I think you should change it. Just add a frame like this somewhere in `demo.tex`.
```TeX
{
\setbeamertemplate{frame footer}{My custom footer}
\begin{frame}[fragile]{Frame footer}
	\themename defines a custom beamer template to add a text to the footer. It can be set via
	\begin{verbatim}\setbeamertemplate{frame footer}{My custom footer}\end{verbatim}
\end{frame}
}
```

davidcortesortuno:
Hi, sorry for the delay. I simplified the code as you suggested. When using `setbeamertemplate` I could not find a way to use the default value (which show the section titles) into the command, which can be done with `metroset` and `custom`, maybe you know more about this. Or I can just simply remove an explanation I made about that in the documentation.

rchurchley:
Thanks for following up on this! To make things as simple as possible, we could even remove support for the syntax `\metroset{footer={custom={My Footer}}}`. That is, we can safely delete the package option from beamerouterthememetropolis.dtx (lines 46–65) and from the documentation. Since it's been a while since you opened this pull request, you'll also have to rebase your changes against master before @matze can merge it in.

This is a great addition to the theme, so thanks again for your work!

davidcortesortuno:
Thanks. I removed those lines and also removed the footer description from the doc, since we no longer use `metroset`

matze:
I rebased your changes and squashed them to a [single commit](https://github.com/matze/mtheme/commit/1b0699893d710a3327eed400c8d59f6f7f5f5348), thanks for your contribution.

PR #172, opened 2016-02-02
innertheme: add noframenumbering for section and titlepage

ronnychevalier:
There is no need to increment the frame number for the titlepage and the
section page.


matze:
> There is no need to increment the frame number for the titlepage and the
section page.

That is true indeed and a regression from a previous version. Thanks for the fix!

PR #174, opened 2016-02-07
Move titleformat options to font theme (and other tidying)

rchurchley:
This PR has two primary purposes: to tidy up the documentation/code style, and to make the component sub-packages of Metropolis more useful on their own. 

1. **Moves all titleformat options to the font theme**. The `titleformat title`, `titleformat subtitle`, `titleformat section` and `titleformat frame` options should, philosophically speaking, be font options, but up until now they were split between the inner and outer themes for implementation reasons. Thanks to a new implementation of `titleformat title` and `titleformat subtitle`, this PR is able to move these options to the font theme and make them independent from the other sub-packages. 

2. **Hyphenates the titleformat options**. Although `titleformat title` et al. are valid PGF keys and work in the `\metroset` macro, the intended use case `\usetheme[titleformat title=smallcaps]{metropolis}` was broken. (It turns out that LaTeX removes all spaces when processing package options.) This PR replaces them with hyphenated versions of the same names (e.g. `titleformat-title`).

3. **Standardizes the 'namespace' of internal macros**. We were previously inconsistent over whether to name an internal macro like `\metropolis@foo` or `\@metropolis@bar`; we should stick to one or the other. Since the `beamer` source code itself uses the first style, this PR removes the leading `@` from internal macros that have it.

4. **Changes the name of `\iffontsexist`** because I kept reading it as "if font sexist". This PR changes it to the less ambiguous `\iffontsavailable`.

5. **Updates documentation**.

This PR also (rather presumptuously) increments the version to 1.1 and updates the date in each `\ProvidesPackage` declaration. Is there a roadmap for an actual 1.1 release and CTAN update?

What do you think about these changes?

### DIFF
--- a/demo/demo.tex
+++ b/demo/demo.tex
@@ -65,7 +65,7 @@
 \end{frame}
 
 {
-\metroset{titleformat-frame=smallcaps}
+\metroset{titleformat frame=smallcaps}
 \begin{frame}{Small caps}
 	This frame uses the \texttt{smallcaps} titleformat.
 
@@ -76,7 +76,7 @@
 }
 
 {
-\metroset{titleformat-frame=allsmallcaps}
+\metroset{titleformat frame=allsmallcaps}
 \begin{frame}{All small caps}
 	This frame uses the \texttt{allsmallcaps} titleformat.
 
@@ -89,7 +89,7 @@
 }
 
 {
-\metroset{titleformat-frame=allcaps}
+\metroset{titleformat frame=allcaps}
 \begin{frame}{All caps}
 	This frame uses the \texttt{allcaps} titleformat.
 
--- a/demo/demo.tex
+++ b/demo/demo.tex
@@ -65,7 +65,7 @@
 \end{frame}
 
 {
-\metroset{titleformat frame=smallcaps}
+\metroset{titleformat-frame=smallcaps}
 \begin{frame}{Small caps}
 	This frame uses the \texttt{smallcaps} titleformat.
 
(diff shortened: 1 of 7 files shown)


matze:
> This PR also (rather presumptuously) increments the version to 1.1 and updates the date in each \ProvidesPackage declaration.

Because the changes of the keys would break existing presentations, I'd rather increment the version to 2.0 according to the [semantic versioning](http://semver.org/) convention.

> Is there a roadmap for an actual 1.1 release and CTAN update?

Not really but I'd handle it how I handle regular software releases. As soon as enough fixes and non-breaking changes pile up, I'd increase the minor version level and release a new version on CTAN.

> What do you think about these changes?

I like those cleanup efforts and see no obstacles in merging them in. What about you @benjamin-weiss?

benjamin-weiss:
> Hyphenates the titleformat options.

Are you sure there is no other solution to fix this. It would obviously be better if we got the version with the spaces to work. As a starting point, this does work:
```TeX
\usetheme[{titleformat frame}=smallcaps]{metropolis}
```
The rest of the changes seem fine to me.

benjamin-weiss:
Maybe defining aliases is a better solution?
```TeX
\pgfkeys{
  /metropolis/outer/titleformatframe/.code=\pgfkeysalso{
     titleformat frame=#1,
  }
}
```

rchurchley:
> Maybe defining aliases is a better solution?

That's a great idea and is now implemented. Keeping the keys with spaces means we can keep the semantic version at 1.x.

benjamin-weiss:
Well then I think @matze can merge this PR.

PR #175, opened 2016-02-12
Re-enable TeX ligatures in XeTeX.

m-oliver:
The font theme on XeLaTeX does not enable the regular TeX ligatures.  This issue is discussed, for example, in 

https://tex.stackexchange.com/questions/37855/fontspec-with-helvetica-breaks-quotes

This one-line patch implements one of the solutions suggested on Stackexchange.

### DIFF
--- a/source/beamerfontthememetropolis.dtx
+++ b/source/beamerfontthememetropolis.dtx
@@ -54,6 +54,14 @@
   \RequirePackage[no-math]{fontspec}
 %    \end{macrocode}
 %
+% Xe\LaTeX{} does not enable regular \TeX{} ligatures so that
+% double-quotes and dashes are not printed correctly.  The following
+% line re-enables traditional \TeX{} ligatures.
+%
+%    \begin{macrocode}
+  \defaultfontfeatures{Ligatures=TeX}
+%    \end{macrocode}
+%
 % \begin{macro}{\checkfont}
 %    Checks if a font is installed; if not, |fontsnotfound| is increased.
 %    \begin{macrocode}


matze:
Seems you are affected by the same problem as reported by me in #154. fontspec does in fact enable ligatures but you probably have an older fontspec version.

@benjamin-weiss: what exactly was speaking against enabling TeX ligatures by default?

benjamin-weiss:
First of all the line should be 
```TeX
\defaultfontfeatures[\rmfamily,\sffamily]{Ligatures=TeX}
```
because we don't want ligatures for the monospaced font.

The downside of adding it to the theme is, that if somebody uses a custom `fontspec.cfg` file the `defaultfontfeatures` settings in it will be overwritten by the theme. And I don't think we should do that if not absolutely necessary.

And yes you are right: it is a default setting now for over 2 years, so @m-oliver should probably update his TeX distribution.

m-oliver:
Thanks for the replies.  I am on Fedora 23 and realize that they are still on TeXlive 2014.  It's easier for me to patch up the theme than to update the distribution manually, especially given that I need this for a collaborative project and cannot expect everybody to update their TeX just for the beamer theme.  But of course you may have different priorities for the master branch, which is fine.  (Personal remark: I like your theme and it's a bit of a shame that it does not run with a fontspec file dated May 23, 2014, which is a bit less than two years.  I have yet to meet anybody messing with their fontspec.cfg, but having older TeX installations around is faily common, for the better or worse...)

benjamin-weiss:
> ... it's a bit of a shame that it does not run with a fontspec file dated May 23, 2014.

Sorry, but I don't see it that way. The theme runs fine with this version (afaik) and it just respects your local fontspec settings. If you want to have ligatures then you just have to add
```TeX
\defaultfontfeatures[\rmfamily,\sffamily]{Ligatures=TeX}
```
to your preamble, like you have to with every other document you typeset on your system using fontspec.

But you are right i mixed up the months and days in the date notation in the change log. So it is not yet a default setting for two years. Version 2.4 was released June 01, 2014. So they unfortunately missed the TeX Live 2014 deadline by some days. 

@matze What is your take on this issue? We already have seen that I am a bit too hostile to old versions. :smirk: Is adding some infos to the "Known Issues" section enough here or should we overwrite the default font features?

m-oliver:
The problem is that adding `\defaultfontfeatures` to the preamble does not help.  It seems to be necessary to patch the font theme, adding this command after loading the fontspec package and before `\setXXXfont`.

benjamin-weiss:
I don't see a problem there. But I could be wrong, as I don't have an old fontspec version at hand right know to test it.

```TeX
\documentclass{beamer}

\usepackage{fontspec}
\defaultfontfeatures[\rmfamily,\sffamily]{Ligatures=TeX}
\usetheme{metropolis}

\begin{document}
\begin{frame}{Ligatures Test}
	\begin{itemize}
		\item fi
		\item fl
	\end{itemize}
\end{frame}
\end{document}
```

m-oliver:
Benjamin Weiss writes:
 > \usepackage{fontspec}
 > \defaultfontfeatures[\rmfamily,\sffamily]{Ligatures=TeX}
 > \usetheme{metropolis}

Sure, that's the next logical step in reverse engineering - I realized
just after I sent the last message.  I still don't think that these
are the details you'd want to expose to the user.  Surely there must
be a way to make sure that the defaults are reasonably set even when
they aren't and stilll get out of the way of users who manually
override their defaults?

benjamin-weiss:
Hmm, maybe we could check for the package version and implement it like this:
```TeX
\@ifpackagelater{fontspec}{2014/06/01}{}{%
    \defaultfontfeatures[\rmfamily,\sffamily]{Ligatures=TeX}
}%
```

matze:
If this works, I'd definitely include it. Better than annoying people stuck with older TeX distributions [like myself ;-)]

XVilka:
Was anything decided about this? More than 3 years passed.

rriemann:
> Hmm, maybe we could check for the package version and implement it like this:
> 
> ```
> \@ifpackagelater{fontspec}{2014/06/01}{}{%
>     \defaultfontfeatures[\rmfamily,\sffamily]{Ligatures=TeX}
> }%
> ```

I don't think this is necessary in 2020. I suggest to close this pull request without merge.

PR #177, opened 2016-02-22
Improve compatibility with other color themes

rchurchley:
Some users may want to override those defined by Metropolis, either by including a stock colour theme or by defining their own colours. However, if their preferred theme sets a background colour for block environments, they will by default not have enough padding:

```latex
\useoutertheme{metropolis}
\useinnertheme{metropolis}
\usefonttheme{metropolis}
\usecolortheme{orchid}
```

Currently, Metropolis uses an option on the inner theme to control the padding on block elements, so the user could manually set `inner/block=transparent` or `inner/block=fill` based on their colour theme. However, this solution is not very intuitive or discoverable.

This pull request solves this inconvenience and fixes a couple other problems involving the use of other color themes:

- Changes the inner theme's definition of block environments so that the `inner/block` option does not need to be set manually. (We use  `\ifbeamercolorempty` to test whether a background colour has been set for `block title` and `block body` and adjust the padding accordingly.)
- Removes the no-longer-used option `inner/block`, simplifying the internal options-handling code. This will not affect any presentations which use, e.g., `\usetheme[block=fill]{metropolis}`, as the option is still passed to the color theme.
- Fixes the known issue with `\plain` slides under the color theme `fly` — the solution handles the case where `palette primary.bg` is undefined.
- Adds documentation describing potential issues when using other color themes with Metropolis.

### DIFF
--- a/doc/metropolistheme.dtx
+++ b/doc/metropolistheme.dtx
@@ -523,7 +523,40 @@ only alphabetic characters and do not require the expansion of any macros.
 
 
 
+\subsection{Interactions with other color themes}
 
+\themename can be used along with any other Beamer color theme, such as
+|crane| or |seahorse|. If you wish to do this, it is usually best to include
+the \themename subpackages individually so the \themename color theme is
+never loaded. Although this will disable some features, including
+|\plain| slides, it will prevent conflicts between the \themename color theme
+and your preferred theme.
+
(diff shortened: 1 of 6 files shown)


matze:
Thanks Ross, changes look good and work for me :+1:

Issue #165, opened 2015-12-14
Frametitle size with parentheses


danjjl:
The vertical size of the frametitle is bigger for titles with parentheses than for titles without.

MWE:

    \documentclass{beamer}
    \usetheme{metropolis}
    \begin{document}

      \begin{frame}{Title with (parentheses)}
      \end{frame}

      \begin{frame}{Title without parentheses}
      \end{frame}
    \end{document}

![screenshot from 2015-12-14 23-53-11](https://cloud.githubusercontent.com/assets/747240/11796697/e7fa1ca0-a2bd-11e5-8928-3fca9ec86b54.png)

This issue looks a lot like https://github.com/matze/mtheme/issues/131.

benjamin-weiss:
You are right this is basically the same problem as #131.

This problem seems to be never-ending with the current implementation using `sep=...`. Maybe the original implementation using `ht=..., dp=..., leftskip=..., rightskip=...` wasn't so bad after all. And simply adjusting `ht=...` and `dp=...` depending on the frame title format should fix #99, too. The only problem I see is, that it will break the support for multiline frame titles again.

rchurchley:
Actually, I think this is the same as #33. The minimum height of the frametitle is set (via struts) to be the maximum height of an alphabetic character. This does not account for the fact that punctuation can be larger than alphabetic characters under `allsmallcaps`.

This issue should be added to the list of unexpected behaviours caused by `allsmallcaps` (as discussed in #138) and can be mitigated by changing the default to `regular` or `smallcaps` (also discussed there).

benjamin-weiss:
@rchurchley I thought about it some more. What do you think about that. We define our strut (in my example for the `regular` case) like that:
```TeX
\renewcommand{\@metropolis@frametitlestrut}{%
  \newdimen\heightOfA
  \setbox0=\hbox{A}
  \heightOfA=\ht0 \advance\heightOfA by \dp0
  \vphantom{\rule[-1.5ex]{1pt}{3ex+\heightOfA}}%
}
```
And change our frame title box definition to this.
```TeX
\begin{beamercolorbox}[%
    wd=\paperwidth,%
    sep=0pt,%
    leftskip=1.5ex,%
    rightskip=1.5ex,%
  ]{frametitle}%
```
Then an **A** is perfectly vertically centered inside the box and every other character will just use the free space above or below, if it needs it.

benjamin-weiss:
[...]

PR #182, opened 2016-02-24
Fix multi-line block titles

rchurchley:
For quite a while, ᴍᴇᴛʀᴏᴘᴏʟɪs has had trouble with block environments with long titles:

screen shot 2016-02-24 at 10 05 00 am

This pull request solves this issue while maintaining the current behaviour for single-line blocks:
screen shot 2016-02-24 at 10 05 55 am
screen shot 2016-02-24 at 10 06 45 am

Technical details:
- Previously, we fixed the height and depth of the `beamercolorbox` in block titles, and added padding to the right and left when `block title.bg` was nonempty. Now, we set `sep` for the `beamercolorbox` and add negative padding to the right and left when `block title.bg` is empty. This is similar to how we fixed multi-line frame titles a while back.
- Just replacing `dp` and `ht` with `sep` wouldn't be enough, though. The padding at the top of a beamer box seems to actually be `sep + parskip`; since ᴍᴇᴛʀᴏᴘᴏʟɪs increases the `\parskip` length, we'd have too much padding at the top if we didn't do something more. The solution is to reset `\parskip` to zero at the beginning of `block begin` and restore the ᴍᴇᴛʀᴏᴘᴏʟɪs `\parskip` length when the block body begins.
- This pull request also adds some wiggle room to the `rightskip` for the block title to produce a good ragged right edge.
- Finally, I added some comments explaining why we can't just use `colsep`, which would be the ideal solution if it actually worked.

### DIFF
--- a/source/beamerinnerthememetropolis.dtx
+++ b/source/beamerinnerthememetropolis.dtx
@@ -325,43 +325,58 @@
 %    the auxiliary macro |\metropolis@block| to define all three templates.
 %
 %    \begin{macrocode}
-\newlength{\metropolis@blocktitleskip}
-\newlength{\metropolis@blockbodyskip}
+\newlength{\metropolis@blocktitleadjustleft}
+\newlength{\metropolis@blocktitleadjustright}
+\newlength{\metropolis@blockbodyadjust}
 \newcommand{\metropolis@block}[1]{
-  \setlength{\parskip}{\metropolis@parskip}
+  \par\vskip\medskipamount
 %    \end{macrocode}
 %
-%    If a background color is defined for the block title or body, a little
-%    bit of padding is added to the corresponding box.
+%    If a background color is defined for the block title or body, we need to
+%    add a little bit of padding to the corresponding box. Ideally, this would
+%    be accomplished by setting |colsep=1ex|, which is intended to add
+%    ``color separation space'' only when the box has a colored background.
+%    Unfortunately, |colsep| also adds this separation if the background color
+%    is inherited, even if the inherited color is actually empty.
+%    (The technical reason for this boils down to the fact that the |\ifx|
+%    directive does not expand macros.)
+%
+%    To achieve the correct spacing for |alertblock|s and |exampleblock|s
+%    as well as for normal blocks, we have to set |sep=1ex| in all cases and
+%    then subtract it from the left and right edges if |block title| has
+%    an empty background. This solution assumes that either all or none of
+%    |block title|, |block title alerted|, and |block title example| have an
+%    empty background, and likewise for |block body|, |block body alerted|,
+%    and |block body example|.
 %
 %    \begin{macrocode}
-  \ifbeamercolorempty[bg]{block title}
-    {\setlength{\metropolis@blocktitleskip}{0ex}}
-    {\setlength{\metropolis@blocktitleskip}{1ex}}
+  \ifbeamercolorempty[bg]{block title}{
+    \setlength{\metropolis@blocktitleadjustleft}{-1ex}
+    \setlength{\metropolis@blocktitleadjustright}{-1ex plus 4em}
+  }{
+    \setlength{\metropolis@blocktitleadjustleft}{0pt}
+    \setlength{\metropolis@blocktitleadjustright}{0pt plus 4em}
+  }
   \ifbeamercolorempty[bg]{block body}
-    {\setlength{\metropolis@blockbodyskip}{0ex}}
-    {\setlength{\metropolis@blockbodyskip}{1ex}}
-  \vspace*{1ex}
-%    \end{macrocode}
-%
-%    Each block environment consists of two (possibly coloured) beamer boxes:
-%    one for the title and one for the body.
-%
-%    \begin{macrocode}
+    {\setlength{\metropolis@blockbodyadjust}{-1ex}}
+    {\setlength{\metropolis@blockbodyadjust}{0ex}}
+  \setlength{\parskip}{0pt}
   \begin{beamercolorbox}[%
-    ht=2.4ex,
-    dp=1ex,
-    leftskip=\metropolis@blocktitleskip,
-    rightskip=\metropolis@blocktitleskip]{block title#1}
-      \usebeamerfont*{block title#1}\insertblocktitle%
+    sep=1ex,
+    leftskip=\metropolis@blocktitleadjustleft,
+    rightskip=\metropolis@blocktitleadjustright]{block title#1}
+      \usebeamerfont*{block title#1}%
+      \insertblocktitle%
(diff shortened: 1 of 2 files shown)


benjamin-weiss:
Seems fine to me, although the solution is bit weird. I would probably prefer a solution the other way around. Set `sep=0pt` by default, add custom struts at the beginning and end (like I proposed [here](https://github.com/matze/mtheme/issues/165) with [this branch](https://github.com/benjamin-weiss/mtheme/commit/047c5ab4412e9863ecf40e40499a8e260b156511) for the frame titles but never submitted the PR) and set `leftskip` and `rightskip` if the background color is *not* empty.

rchurchley:
Coming from a web/CSS background, setting the desired padding and compensating for a special case with a negative margin was actually the more intuitive solution to me.

I prefer using Beamer's higher-level options where possible rather than rolling our own solutions with struts. Using Beamer's mechanisms mean that our bugs will at least be consistent with other themes. I'm sure that a strut-based solution is possible, but I'm not sure how well I could debug it, or to anticipate the relevant edge cases where our solution would differ from other themes. (3+ line titles? Accented capital letters?)

In an ideal world, of course, we'd just use `colsep` and not have any colour-conditional code at all. I suppose that's another advantage of the `sep` solution — if Beamer fixes the bug with `colsep` and inherited empty colours, it will be very easy to take advantage of that.

benjamin-weiss:
I have done my fair share of CSS too and that is the reason why the `sep` or `colsep` parameter looks so strange to me. Consider the following example in HTML/CSS:

```HTML
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Padding Test</title>
 </head>
<body>
<p style="padding: 1ex;">
<div style="padding:1ex; background:#E6E8E8; font-size: 30pt; font-family: 'Fira Sans'; display: inline; margin-left:1ex;">
ALL CAPS
</div>
<div style="padding:1ex; background:#D0D4D5; font-size: 30pt; font-family: 'Fira Sans'; display: inline">
Regular
</div>
</p>
</body>
```
You can see that whatever text you write inside the `div`, it has always the same height. That is not the case with `beamercolorbox`es. A `beamercolorbox` with the same font size and the same `sep` can have a different height. I illustrated this earlier with this image.

1c7afd4e-2ebc-11e5-96a6-a7d19d75b1df

With CSS and a fixed padding setting the distances, a_2, c and b are the same. My custom solution does exactly this.

rchurchley:
Ah, I see my misunderstanding. By "the solution is bit weird", I thought you were objecting on code readability grounds. You bring up a very valid point about box heights. 

I originally thought that maintaining the height of block titles was not as important as it is for the frametitles, as blocks are usually not positioned in a way that the eye naturally compares the heights. But you're absolutely right; there are common use cases where my current solution looks bad:

screenshot

Implementing the padding entirely with struts, as you did in #165, would be one way to go. However, since we don't have to satisfy the constraints of an `allsmallcaps` option for the block titles, I don't think we need to go quite so far. It should be fine to keep `sep` for the padding inside the box, and just use `\vphantom{ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghikjlmnopqrstuvwxyz()}` to ensure a consistent line height.

screenshot-1

What do you think?

matze:
Hmm, although the latter is objectively more consistent the (relatively) larger title backgrounds seem to emphasize them even more and de-emphasize the actual content. On the other hand I won't use filled blocks anyway, so this is good for me. I'll wait for @benjamin-weiss' feedback before merging this in.

benjamin-weiss:
The screenshot is a bit small and I don't have time right now to check it myself. But I also think the block title block is a bit to big. That is one of the reasons I implemented my "overcomplicated" custom struts. Right now a_1, c and "b_2" are the same length. I will check it out thoroughly tonight (in around 10 hours).

rchurchley:
@matze Actually, that's another good point. Without making the struts conditional on the background, this solution also affects transparent blocks, adding what I feel is too much space between the block title and block body (see below left)

untitled

Unless we make our struts conditional on the block background, we'll also have to be careful to test how our solutions affect transparent blocks.

rchurchley:
I've tweaked the numbers a little bit in the hopes of getting something a little closer to what we'd like, based partly on Beamer's default measurements (`colsep=0.75ex`) and partly on a rough calculation that suggests parentheses contribute about `0.25ex` to each of the top and bottom padding.

screen shot 2016-02-26 at 1 18 54 am
screen shot 2016-02-26 at 1 18 46 am

My latest version no longer forces transparent and filled blocks to have exactly the same spacing, so we can optimize them independently. (Apart from the demo presentation, it is unlikely that filled and transparent block environments will appear on the same slide.)

screen shot 2016-02-26 at 1 19 32 am
screen shot 2016-02-26 at 1 19 19 am

I would appreciate you two giving this a closer look and sharing your opinions and suggestions!

benjamin-weiss:
I still don't like the behavior of the `beamercolorbox` `sep` parameter, but with this small value it doesn't really stand out. So although I would prefer an implementation similar to my frame title struts I'm fine with it here. Overall it is a big improvement to the previous implementation, so I would just merge it right now. The only thing you maybe could do is to reduce the number of hardcoded lengths.

But we maybe should get back to this issue at some point and try to figure out a real solution or better workaround for the `ifbeamercolorempty` bug. But right know I also don't see a better way.

PS: Just a minor issue as probably hardly anybody is ever gonna do it, but setting only the body background to empty also looks strange.
```TeX
\documentclass{beamer}
\usetheme[block=fill]{metropolis}
\begin{document}
\begin{frame}{Blocks}
  \setbeamercolor{block body}{bg=}
  \begin{block}{Default}
     Block content.
  \end{block}
\end{frame}
\end{document}
```

rchurchley:
> The only thing you maybe could do is to reduce the number of hardcoded lengths.

I've refactored the code again to make the lengths less hacky. and added comments so it's a bit clearer what's going on! I don't see any regressions on the presentation I've been using to test, but I'd appreciate one last scan by you guys and then we can merge it.

> Just a minor issue as probably hardly anybody is ever gonna do it, but setting only the body background to empty also looks strange.

I think that's probably unavoidable. Fortunately, anybody who wanted to do that for some reason could instead always use `\setbeamercolor{block body}{bg=normal text.bg}` to have a transparent-looking background with fill-compatible spacing.

benjamin-weiss:
On last thing. You could discard your changes to the outer theme that our two PRs do not have any conflicts.

rchurchley:
Good point, and now done. I think this should be ready to merge now?

matze:
Sorry for the late reply, have been on a short vacation. I think this and #183 are ready for a merge. Thanks!

PR #183, opened 2016-02-26
new frame title struts – fixes #165

### LINKED ISSUE #165: Frametitle size with parentheses

benjamin-weiss:
As my original branch was too outdated I had to reimplement this. As @rchurchley moved all the title format settings to the font theme the struts couldn't be declared alongside the title format settings anymore without keeping the outer theme and the font theme independent from each other.

So I added a check if `metropolis@frametitleformat` is defined. If this is the case we use it in the calculation and otherwise not.

Probably somebody should check it again if everything is working as expected, but I'm confident it does.

### DIFF
--- a/source/beamerouterthememetropolis.dtx
+++ b/source/beamerouterthememetropolis.dtx
@@ -159,11 +159,7 @@
 \newcommand{\metropolis@frametitlestrut@start}{
   \rule{0pt}{\metropolis@frametitle@padding +%
     \totalheightof{%
-      \ifcsname metropolis@frametitleformat\endcsname%
-        \metropolis@frametitleformat X%
-      \else%
-        X%
-      \fi%
+      \ifcsdef{metropolis@frametitleformat}{\metropolis@frametitleformat X}{X}%
     }%
   }%
 }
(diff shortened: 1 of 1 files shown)


rchurchley:
Looks right to me, although I should also probably test this later too.

Suggestion: would it be preferable to do the conditional with `etoolbox`?
```latex
\ifcsdef{metropolis@frametitleformat}{\metropolis@frametitleformat X}{X}
```
I don't know of any appreciable difference between the two; it's just another possible option if you prefer the style.

benjamin-weiss:
> I don't know of any appreciable difference between the two

As far as I understand it they are basically identical and it is just about readability. But I think it makes sense to use the `etoolbox` macro so I changed it.

benjamin-weiss:
I was just having a last look at the demo slides and am now wondering if we should slightly increase the `\metropolis@frametitle@padding`. As my implementation results in a smaller block, compensating it is probably a good idea. Maybe 
```TeX
\setlength{\metropolis@frametitle@padding}{2ex}
```
is a better choice?

rchurchley:
I agree, the frame titles do look better if the padding is set to 2ex. Other than that, everything looks good in my tests!

Issue #11, opened 2014-12-02
Make plain frame command another instance of frame?


talexand:
If the `\plain{[title]}{body}` command for dark frames matched Beamer frame syntax, it would allow Beamer exporters like [emacs org-mode](https://www.gnu.org/software/emacs/manual/html_node/org/Beamer-export.html#fnd-1) to reference it using existing templates.

I can come up with something but I need a little bit of a sketch on defining frames in LaTeX style files as I've never done it before.

rchurchley:
What is meant by "Beamer frame syntax"? If you just want

```latex
\begin{plainframe}{...title...}
    ...
\end{plainframe}
```

we should be able to accomplish this using `\newenvironment` instead of `\newcommand`. Or am I missing something?

rchurchley:
I've finally done some research into this long-standing issue. I believe I originally misunderstood @talexand's request; by "matching frame syntax", he means implementing the standout plain slides as an optional argument to `frame`.

```latex
\begin{frame}[standout]
  Questions?
\end{frame}
```
(Obviously, we couldn't use `plain` as the argument, as this already exists in Beamer. I've called the option `standout` for now since I couldn't think of anything better, but I'd welcome other ideas for what to call it.)

As I understand it, doing this would allow users of emacs org-mode to generate the above frame using something like the following input. (Note: I have not actually tested it.)

```
* A standout slide                    :B_ignoreheading:
:PROPERTIES:
:BEAMER_opt: standout
:END:
    Questions?
```

## Advantages

Moving to an optional frame argument offers a couple mild advantages in addition to emacs org-mode compatibility.

- The syntax is more consistent with other frames.
- It could make it very slightly easier to adapt a presentation to or from ᴍᴇᴛʀᴏᴘᴏʟɪs.
- We could implement the new option in the inner theme, making it possible to use independently of the font and color themes. (For some reason, I feel more comfortable defining a new frame option in the inner theme than a new command.)

## Disadvantages

The main disadvantage of this proposal is that I don't see a way of implementing the `allsmallcaps` and `allcaps` options of `titleformat plain`. This can be mitigated by keeping `\plain{...}` as a shortcut for `\begin{frame}[standout]{#1}\metropolis@plaintitleformat{#2}\end{frame}`

## Implementation

[...]

PR #189, opened 2016-03-05
Use frame key for standout slides, move to inner theme

rchurchley:
As discussed in #11, I have reimplemented the `\plain` command for focused, colour-inverted slides as an optional key for Beamer frames. The new syntax is:

```latex
\begin{frame}[standout]
  Questions?
\end{frame}
```

This change does not affect older presentations which use `\plain{...}`, although as per @matze's suggestion, `\plain` now issues a warning that it "may be deprecated" in a future version and `[standout]` should be preferred.

### DIFF
--- a/demo/demo.tex
+++ b/demo/demo.tex
@@ -301,7 +301,9 @@ or show \textbf{bold} results.\end{verbatim}
 
 \end{frame}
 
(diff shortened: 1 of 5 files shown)


matze:
Great, thanks!

PR #190, opened 2016-03-05
Add support for subsections

rchurchley:
Although I haven't generally used subsections in my beamer presentations, I understand that they can be an essential feature when giving long tutorials or multi-day workshops, for example. @eddelbuettel asked us to implement some form of subsectioning in #53, and this pull request is making good on my promise nearly a year ago to do so. :)

Without duplicating any code, this pull request defines two `subsection page` templates matching the `section page`s, containing both the section and subsection title.

`simple`
screen shot 2016-03-04 at 9 46 54 pm

`progressbar`
screen shot 2016-03-04 at 11 31 11 pm

These templates are not enabled by default; as @matze pointed out in #53,

> there is a tendency to break up a presentation into smaller topics if a theme encourages that (e.g. showing little dots and upcoming titles).... A more organic flow probably helps [the audience stay] awake than a overly hierarchic structure.

To encourage presenters to use subsections only as a tool to orient the audience during a long presentation, we require them to use the new inner theme option `subsectionpage` if they want to turn on subsection slides. 

The possible values of `subsectionpage` are `none`, `simple`, and `progressbar`, matching those of `sectionpage`. In fact, the templates for the section and subsection pages are tied together so they will always match if both are enabled. They can, of course, be enabled and disabled independently.

### DIFF
--- a/doc/metropolistheme.dtx
+++ b/doc/metropolistheme.dtx
@@ -356,6 +356,12 @@ The list of options is structured as shown in the following example.
   disables the section page.
 }
 
+\DescribeOption{subsectionpage}{none, simple, progressbar}{none}{
+  Optionally adds a slide at the start of each subsection. If enabled with
+  the |simple| or |progressbar| options, the style of the |section page| will
+  be updated to match the style of the |subsection page|.
+}
+
 
 \subsubsection{Outer theme}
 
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Love it :)

matze:
That indeed looks a clean solution to that old age issue ;-)

PR #194, opened 2016-03-09
Allow user to define empty (e.g.) alertblock background

rchurchley:
When I use filled block environments, I do so to structurally separate my main results — theorems, lemmas, questions, etc. — from the discussion surrounding them. I consider examples to be "discussion", so I prefer not to have them filled.

Using ᴍᴇᴛʀᴏᴘᴏʟɪs, the obvious thing to do is

```latex
\metroset{block=fill}
\setbeamercolor{block title example}{bg=}
\setbeamercolor{block body example}{bg=}
```

However, this produces incorrect spacing to the left of the example block:

screen shot 2016-03-08 at 6 58 21 pm

This pull requests fixes this bug.

screen shot 2016-03-08 at 6 57 16 pm

(By the way, @matze, since I now have three pull requests affecting the inner theme open, please let me know if you need me to rebase each against the others. I don't think there should be a problem, since they address different parts of the file, but just in case.)

### DIFF
--- a/source/beamerinnerthememetropolis.dtx
+++ b/source/beamerinnerthememetropolis.dtx
@@ -355,10 +355,13 @@
 %    |block title alerted|, and |block title example| have an empty
 %    background).
 %
-%    If the |block title| background is empty, we just need to set a rightskip
-%    for a nice ragged-right block title.
+%    If the |block title| background is empty, or the user has explicitly
+%    removed the background from (e.g.) |block title alerted|, we just need to
+%    set a rightskip for a nice ragged-right block title.
 %
 %    \begin{macrocode}
+  \ifbeamercolorempty[bg]{block title#1}{%
+    \begin{beamercolorbox}[rightskip=0pt plus 4em]{block title#1}}{%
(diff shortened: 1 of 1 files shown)


matze:
Thanks for the bug fix. One thing that I noticed lately is that the padding of various elements was a lot less than before. I [increased it again for the frame title](https://github.com/matze/mtheme/commit/f31c772a8fa2dc2f17fd1473911a702b9969688b) but start noticing that also with the blocks. I think a little more white space could alleviate that cramped feeling in my opinion, what do you think?

Concerning your comment: As long as I get a green merge button and you do not merge or cherry-pick commits from master in between, everything is fine.

rchurchley:
> I think a little more white space could alleviate that cramped feeling in my opinion, what do you think?

Go for it! The only reason the current block sep is `0.75ex` is because I borrowed from `beamerinnerthemedefault` when reimplementing `block begin`. The padding is set with `\metropolis@blocksep` in the inner theme so it should be easy to tweak.

Issue #193, opened 2016-03-08
Block background color


alcarazr:
Hi,

if I am not mistaken, currently all three block types share the same background color, which is based on the `normal text` color (both `fg` and `bg`).

Have you considered setting the background color depending on the block type (depending on the `alerted` and `example` `fg` colors, respectively)?

Thanks,

benjamin-weiss:
> if I am not mistaken, currently all three block types share the same background color

You are not mistaken, that's right.

> Have you considered setting the background color depending on the block type

Do you mean something like this?

![demo verschoben](https://cloud.githubusercontent.com/assets/9158719/13617757/a3586682-e580-11e5-8419-7d7c492c1aa5.png)

Probably not, because that doesn't look good at all (imho). Could you make an example what you mean? You just have to set the colors manually after loading the theme with
```TeX
\setbeamercolor{block title alerted}{ ... }
\setbeamercolor{block body alerted}{ ... }
```
for the alerted block and can then post a picture here.

But either way, I think it is a good decision to use the same background color for all blocks. In my opinion too much color is too distracting.

rchurchley:
>> Have you considered setting the background color depending on the block type?

> Do you mean something like this?

I expect @alcarazr is proposing something along the lines of the Beamer colour themes `crane`, `orchid`, and `rose`:

![screen shot 2016-03-08 at 5 59 40 pm](https://cloud.githubusercontent.com/assets/1131743/13623028/4dab64a0-e558-11e5-81e4-7d28aba0607e.png)

I think this and the current design look equally good and do not have strong opinions on the matter. Having more colour can be distracting sometimes (I tend to feel this way about example blocks) but it can also help focus the audience's attention to an important part of the slide (most common with alert blocks).

I would not object to either changing it or keeping the same. But either way, it's a design choice worth considering. (Thanks for the suggestion, @alcarazr!)

@matze, @benjamin-weiss, what do you think?

alcarazr:
A full color conversion would actually look more like this (code below image):

![blocks](https://cloud.githubusercontent.com/assets/9049533/13628817/743538ee-e5d5-11e5-9bdb-5f6618c32d57.jpg)

```TeX
\documentclass{beamer}

\usetheme{metropolis}

% Colors based on Paul Tol's recommendations for marking text (https://personal.sron.nl/~pault/#textmarking)

\definecolor{BlueTOL}{HTML}{222255}
\definecolor{BrownTOL}{HTML}{666633}
[...]

Issue #191, opened 2016-03-06
White used as font colour if notes enabled


andreskytt:
When building the following MVE:

```
\documentclass{beamer}
\usepackage{pgfpages}
\setbeameroption{show notes on second screen=right} 

\usetheme{metropolis}           % Use metropolis theme
\title{A minimal example}
\date{\today}
\author{Matthias Vogelgesang}
\institute{Centre for Modern Beamer Themes}
\begin{document}
  \maketitle
  \section{First Section}
  \begin{frame}{First Frame}
    Hello, world!
  \end{frame}

\end{document}

```
The words "Hello, world!" are rendered in white instead of the regular colour. The result is attached.
[metropolisdemo.pdf](https://github.com/matze/mtheme/files/160293/metropolisdemo.pdf)

benjamin-weiss:
Have a look at this TeX.SE question: http://tex.stackexchange.com/questions/288408/beamer-show-notes-on-second-screen-with-xelatex-and-atbeginsection

andreskytt:
This does indeed fix the font colour problem but messes up formatting for section titles, references, etc.
```
\documentclass{beamer}
\usepackage{pgfpages}
\setbeameroption{show notes on second screen=right} 

\usetheme{metropolis}           % Use metropolis theme
\title{A minimal example}
\date{\today}
\author{Matthias Vogelgesang}
\institute{Centre for Modern Beamer Themes}


\AtBeginSection{
    \frame{\insertsectionhead}
}

\makeatletter 
\def\beamer@framenotesbegin{% at beginning of slide
     \usebeamercolor[fg]{normal text}
      \gdef\beamer@noteitems{}% 
      \gdef\beamer@notes{}% 
}
\makeatother

\begin{document}
  \maketitle
  \section{First Section}
  \begin{frame}{First Frame}
    Hello, world!
  \end{frame}

\end{document}
```

[metropolisdemo.pdf](https://github.com/matze/mtheme/files/160394/metropolisdemo.pdf)

My LaTeX knowledge is limited but I'll try to have a stab at creating a pull request for fixing this.

benjamin-weiss:
I think you misunderstood something here. We use
```TeX
\AtBeginSection{...}
```
in the metropolis theme to create the section frames. That's why you are affected of this beamer bug if you are using the metropolis theme. And
```TeX
\AtBeginSection{
    \frame{\insertsectionhead}
}
```
overwrites the metropolis section frames. To fix the bug you only need
```TeX
[...]

Issue #180, opened 2016-02-23
Placement of institute on title slide


twsh:
This is a very subjective issue, so I make the following suggestion tentatively. To me it seems that the institute belongs conceptually with the author. Currently it appears a long way away. I think it would be better appearing immediately after with nothing intervening.

benjamin-weiss:
In my opinion the page layout doesn't look very good if the order of date and institute is changed. Here is quick sketch.

![metro verschoben](https://cloud.githubusercontent.com/assets/9158719/13264533/e93fa204-da6e-11e5-90ec-2784c0b42e5a.png)

twsh:
I prefer it that way.

rchurchley:
It's definitely a subjective issue, so I'm not sure whether we're likely to reach a consensus. For what it's worth, I quite like @twsh's suggestion of grouping the author and institute together.

I think we can do it elegantly if we adjust the spacing and slightly mute the colour of the institute:

 
![screen shot 2016-02-23 at 1 04 03 pm](https://cloud.githubusercontent.com/assets/1131743/13266655/f0872dfc-da2d-11e5-8a70-607dc73a7c24.png)

matze:
This is indeed a very subjective issue which is why we should postpone it until we can come up with a flexible solution that satisfies everyone. For now, I'd say, it looks "good enough".

nomeata:
I’m about to start `mtheme` and this was my first stumbling step.

I worked (hacked) around it this way:
```
\title{Actual title}
\subtitle{Conference name\\Date of talk}
\author{Joachim Breitner}
\date{}% needs to be empty
\institute{Author’s institute}
```
This groups the date with the talk, and the institute with the author.

notmatthancock:
Combining @rchurchley 's suggestion with @nomeata 's to give a lighter-grayish-teal subtitle and date:
```
\documentclass{beamer}
\usepackage{xcolor}
\definecolor{LightGray}{rgb}{0.45,0.5,0.45}
\usetheme{metropolis}
\title{A correlation study between burrito girth and tortilla fail point: how many beans is too many?}
\subtitle{\vspace{5pt} \textcolor{LightGray}{Burrito Con \\ Jan 1, 2048}}
\date{}
\author{Author 1, Author 2}
\institute{Institute for Advanced Burrito Studies}
\begin{document}
  \maketitle
\end{document}
```
![](http://i.imgur.com/UAg7SBR.png)

Issue #178, opened 2016-02-22
Restructure pgfplot theme


matze:
As a follow-up to #171, we should restructure the pgfplots styles and decouple color from structural properties.

rchurchley:
As a bit of a radical suggestion: should we remove the pgfplots theme from the ᴍᴇᴛʀᴏᴘᴏʟɪs project and release it as its own separate package? The Tol color theme is a great way to make readable charts with `pgfplots`, but at this point it's almost entirely independent from the ᴍᴇᴛʀᴏᴘᴏʟɪs beamer theme. I wonder if it would be better served by having its own repository.

## Proposal

- Add a warning to `pgfplotsthemetol` for the v1.1 release of ᴍᴇᴛʀᴏᴘᴏʟɪs, indicating that it would be removed in a future version and need to be installed separately.
- Update `beamerthememetropolis.sty` to load `pgfplotsthemetol` only if it is installed on the user's system.
- Freeze development of the version of `pgfplotstheme.sty` included in the ᴍᴇᴛʀᴏᴘᴏʟɪs bundle.
- Create the new GitHub repository *matze/pgfplotstheme-tol*.
- Continue development on `pgfplotsthemetol` in the new repository.
- Have @matze release `pgfplotsthemetol` as its own separate package on CTAN when the new version is ready.
- At the same time, release a v2.0 version of ᴍᴇᴛʀᴏᴘᴏʟɪs on CTAN which removes `pgfplotsthemetol.sty` from the bundle.


## Advantages

- The PGFPlots theme does not depend at all on Beamer and can be used to style charts in articles, books, etc. Releasing `pgfplotsthemetol` as its own package helps communicate this fact, and makes it more visible to users who want a package for this use case.
- A separate manual for `pgfplotsthemetol` would be more useful for all users, especially those who want to use `pgfplotsthemetol` in non-Beamer documents.
- Creating a new repository gives `pgfplotsthemetol` support requests its own issue tracker, which helps ᴍᴇᴛʀᴏᴘᴏʟɪs development stay organized.
- It makes sense philosophically to spin off the PGFPlots theme into its own project. Its code is entirely independent from the other parts of the bundle, and we seem to be developing it separately anyways.


## Possible disadvantages

[...]

PR #195, opened 2016-03-09
Documentation updates

rchurchley:
This pull request makes a few tweaks to the manual. Most notably, it:

- Adds the Beamer/XeLaTeX bug reported in #191 to the Known Issues section, together with the workaround suggested by @benjamin-weiss.
- Recommends using `\setsansfont[BoldFont={Fira Sans SemiBold}]{Fira Sans Book}` under certain presentation conditions in the Font Customization section, as suggested by @ChipmunkMath in #75.
- Removes the link to a Debian package that no longer exists.
- Updates the date of all packages that have been changed since v1.0.

Assuming this can be merged successfully (along with #189, #190, and #194), I think we're just about ready for v1.1. If my count is correct, four issues will remain outstanding and we have decided to postpone each of them to a future release:

#117 — postponed by the issue author.
#178 — postponed for pgfplots to include Paul Tol's SRON colours by default.
#180, #193 — postponed for a longer discussion about the design questions involved.

There will also be two open pull requests. Either of them could be included in the v1.1 release if they are made ready, but they are not urgent and could be postponed if needed.

#100 — could be included if @benjamin-weiss wants to rebase.
#175 — could be included if @m-oliver or @matze implement @benjamin-weiss's suggestion.


matze:
I will merge this one right before the release, in case more documentation related changes need to be added. Thanks so far!

rchurchley:
Sounds good to me!

matze:
I think we are ready :cake:

PR #200, opened 2016-03-24
Fix the "Overfull \hbox (0.4pt too wide)" bug when using progress bar.

wagermn:
When using with the options progressbar, there will be a warning like:
> Overfull \hbox (0.4pt too wide) in paragraph at lines ......

this is because that tikz has a default line width of 0.4pt, just fill the rectangle will solve it.


matze:
Great fix for this longstanding annoyance, thanks!

Issue #202, opened 2016-03-29
One line text title has a background with  a two-line title height


xuhdev:
The following document demonstrates that the first frame title black background occupies two lines while the text only uses one line. The second frame is a comparison of the height of the title black background.

```latex
\documentclass[12pt]{beamer}

\usetheme[sectionpage=none]{metropolis}

\begin{document}

\begin{frame}{AAAAA AAAAA AAAAA AAAAA AAAAA AAAAA}
\end{frame}
\begin{frame}{AAAAA AAAAA AAAAA AAAAA AAAAA}
\end{frame}
\end{document}
```

xuhdev:
Here is the [output pdf](https://github.com/matze/mtheme/files/193138/presentation.pdf).

benjamin-weiss:
Until this bug is fixed in the theme add

```TeX
\makeatletter
\setbeamertemplate{frametitle}{%
  \nointerlineskip%
  \begin{beamercolorbox}[%
      wd=\paperwidth,%
      sep=0pt,%
      leftskip=\metropolis@frametitle@padding,%
      rightskip=\metropolis@frametitle@padding,%
    ]{frametitle}%
\metropolis@frametitlestrut@start\insertframetitle\nolinebreak\metropolis@frametitlestrut@end%
  \end{beamercolorbox}%
}
\makeatother
```
to your preamble.

rchurchley:
@benjamin-weiss Looks like adding that `\nolinebreak` before the strut should be the solution to fix the theme as well. Assuming it doesn't throw an error with an empty or whitespace-only frame title, do you want to submit that one-line fix as a pull request?

benjamin-weiss:
@rchurchley Yes it is the proper fix for this issue and it does not throw an error with an empty or whitespace-only frame title. I just knew the solution and quickly posted it. And actually I should not procrastinate right now by working on metropolis at all, but rather work on my master thesis. :smirk: So feel free to make a PR if you have the time.

PR #204, opened 2016-03-30
Fix empty second line in certain frametitles

rchurchley:
In #202, @xuhdev reported a problem in certain frametitles. If the width of the frametitle is almost exactly `\linewidth`, then `\metropolis@frametitlestrut@end` could be pushed onto the next line, producing an extra blank line in the frametitle. This PR implements @benjamin-weiss's proposed fix for the problem.


benjamin-weiss:
Thanks!

matze:
Looks good :+1:

PR #219, opened 2016-08-10
update unzip filenames

arijitlaik:
the unzip command used filenames from the previous urls the new urls have different file names, this fixes the script


matze:
Thanks!

PR #231, opened 2016-10-21
Fix make uninstall

ntadej:
The double quotes in make uninstall prevented `rm` from working properly at least on OS X

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ install: $(PACKAGE_STY) $(DOC_PDF)
 	@cp $(DOC_PDF) $(DOC_DIR)
 
 uninstall:
-	@rm -f "$(addprefix $(INSTALL_DIR)/, $(PACKAGE_STY))"
+	@rm -f $(addprefix $(INSTALL_DIR)/, $(PACKAGE_STY))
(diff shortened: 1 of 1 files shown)


benjamin-weiss:
Why do you want to uninstall it? Metropolis is awesome. :smile:

matze:
Besides the reason given by @benjamin-weiss, we should look for another solution. The quotes were introduced because passing directory name with spaces in them might delete unrelated directories.

ntadej:
@benjamin-weiss I made a personal fork with some changed colors and defaults, so I renamed it and wanted to test if makefile still works 😊

@matze The current uninstall does not work. I can try iterating over the files and appending path to each one, but I'm not very fluent in make.

rchurchley:
What if we just add quotes to the problematic directory strings? That is, instead of adding quotes to each command in `make uninstall`, we define
```make
DESTDIR ?= "$(shell kpsewhich -var-value=TEXMFHOME)"
```
instead of 
```make
DESTDIR ?= $(shell kpsewhich -var-value=TEXMFHOME)
```
I think this should properly escape the things that need to be escaped without breaking the uninstall script.

Scutua (approved):

PR #238, opened 2016-11-06
fixes #237: Allow to define the width of the progress bar

### LINKED ISSUE #237: Allow to define the width of the progress bar
I propose that the width of the progress bar be selectable by the user via a simple command / option. I have discovered that the current progress bar is hardly visible when the projector used is not of very high resolution, so a wider line would better serve the purpose of the progress bar.

benjamin-weiss:
I don’t think we should provide an option to solve this. So this PR only introduces internal variables for the line width of the progress bar and the title separator, namely:

- `\metropolis@titleseparator@linewidth`
- `\metropolis@progressonsectionpage@linewidth`
- `\metropolis@progressinheadfoot@linewidth`

### DIFF
--- a/source/beamerinnerthememetropolis.dtx
+++ b/source/beamerinnerthememetropolis.dtx
@@ -196,9 +196,11 @@
 %   Template to set the title graphic in a zero-height box. (It won't
 %   change the position of other elements.)
 %    \begin{macrocode}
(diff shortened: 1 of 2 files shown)


matze:
Looks good, thanks!

Issue #239, opened 2016-11-08
clash on fontspec


nbecker:
If tex file contains:
\usepackage{fontspec}
I get:

(/home/nbecker/texmf/tex/latex/metropolis/beamerfontthememetropolis.sty
Package: beamerfontthememetropolis 2016/03/14 Metropolis font theme

! LaTeX Error: Option clash for package fontspec.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.91 }
    
? 
! Emergency stop.
 ...                                              
                                                  
l.91 }
    
The package fontspec has already been loaded with options:
  []
There has now been an attempt to load it with options
  [no-math]
Adding the global options:
  ,no-math
to your \documentclass declaration may fix this.
Try typing  <return>  to proceed.

zhouyisu:
This is because Metropolis already loads fontspec in the preamble. If you want to change font, you can edit beamerfontthememetropolis.sty file.

benjamin-weiss:
@zhouyisu is not quite right. Yes, **metropolis** loads the `fontspec` package (if XeTeX or LuaTeX is used to compile). But the actual problem is, that it uses the `no-math` option, hence the option clash. So for example the following MWE compiles just fine:
```TeX
\documentclass{beamer}
\usetheme{metropolis}
\usepackage[no-math]{fontspec}  % no option clash
\begin{document}
  \begin{frame}{A Frame}
    Hello world!
  \end{frame}
\end{document}
```
But it is usually not necessary to load the `fontspec` package. If you really want to use `fontspec` without the `no-math` option and want to regain full control over the fonts, you could only load the inner, outer and color theme, like this:
```TeX
\documentclass{beamer}
\usepackage{fontspec}
\useinnertheme{metropolis}
\useoutertheme{metropolis}
\usecolortheme{metropolis}
\begin{document}
  \begin{frame}{A Frame}
    Hello world!
  \end{frame}
\end{document}
```
But if @zhouyisu is right and you just wants to change the fonts, you should not follow his advice to change the `beamerfontthememetropolis.sty` file. You should follow the instructions in the manual in section 3.3 and change the fonts after loading metropolis with the standard font selection commands of the `fontspec` package. Like this:
```TeX
\documentclass{beamer}
\usetheme{metropolis}
\setsansfont{Ubuntu}      % set main font
\setmonofont{Ubuntu Mono} % set mono spaced font
\begin{document}
[...]

PR #243, opened 2016-11-20
Fix fontspec option clash

jGleitz:
I stumbled on #239 myself. I could fix it by only loading `fontspec` if it isn’t loaded yet. This solutions has no negative effects from my point of view, but fixes option clashes in some cases.

Note: I was not able to test if the `no-math` option is still passed in correctly, as I can’t produce a case in which it’s needed.

### DIFF
--- a/source/beamerfontthememetropolis.dtx
+++ b/source/beamerfontthememetropolis.dtx
@@ -45,7 +45,11 @@
 %
 %    \begin{macrocode}
 \ifboolexpr{bool {xetex} or bool {luatex}}{
-  \RequirePackage[no-math]{fontspec}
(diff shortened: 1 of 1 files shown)


matze:
Looks good to me, thanks :+1:

nbecker:
This still has the effect of passing no-math to fontspec unconditionally.  Does mtheme actually require no-math?  Why?

nbecker:
so far with a little experimentation, I find that without no-math, \Sigma
is changed, and I think it's a bit too bold (\Sigma is one that is listed
in the log that fontspec changed).  So from very minimal experiment so far,
it looks like no-math looks better.

Issue #224, opened 2016-09-29
Option for an even lighter color scheme


rpls:
In most situations the theme works very well, but I've recently held a presentation with sub-optimal lighting and an older projector which resulted in barely legible slides due to bad contrast. Using a while background helped. An option for this would be nice =) Maybe named "highcontrast".

matze:
This is related to #208 I guess. However, I doubt just making the background pure white would've helped. A darker and thicker font would probably be a better solution, however making this an option will quickly get out ouf hand. I am actually thinking about distributing new color and font schemes with metropolis, e.g. ones that are focused on readability.

matze:
A new high contrast color theme has been added with #244.

Issue #208, opened 2016-04-28
Default color theme


alcarazr:
Hi,

I am loving `mtheme` and here I would like to express my gratitude for all the work put on it. Just a comment regarding the default color theme. It seems that the default `orange` and `green` used for `alerted` and `example` text, respectively, do not pass the contrast test according to the [WCAG Guidelines](http://www.w3.org/TR/WCAG20/#visual-audio-contrast), when the background is either the default `black!2` or pure `white` (as [in the documentation](http://osl.ugr.es/CTAN/macros/latex/contrib/beamer-contrib/themes/metropolis/doc/metropolistheme.pdf) for example).

The project [Colorable](http://jxnblk.com/colorable/) (demo [here](http://jxnblk.com/colorable/demos/text/?background=%23fafafa&foreground=%23EB811B)) allows for getting contrast values for every possible combination and includes pass/fail scores according to the [WCAG](http://www.w3.org/TR/WCAG20/#visual-audio-contrast). A safe color combination should have **Level AAA**. As you can check with the [Colorable demo](http://jxnblk.com/colorable/demos/text/?background=%23fafafa&foreground=%23EB811B), when using the _light_ color theme, the `alerted`and `example`colors fail, whereas they have a greater contrast (but still not AAA) when using the _dark_ theme.

Of course the user can change these colors, but I think it would be nice if the default color theme used a _safer_ color combination. Is this something worth considering?

Thanks,

Rodrigo

matze:
First of all: thanks for bringing up that topic. I agree with you that we have to do something about it. I also thought the orange link color in the docs look a bit odd probably caused by this issue.

> It seems that the default orange and green used for alerted and example text, respectively, do not pass the contrast test according to the WCAG Guidelines, when the background is either the default black!2 or pure white (as in the documentation for example).

As far as I can see, it is impossible to have good contrast with that particular orange-to-reddish color on any kind of background (one can see this nicely in the matrix where there is a similar color). However, with 100% saturation and slightly less lightness we can get the orange at least to AA Large (3.19). Lowering lightness further to achieve AAA destroys the color for me though.

> Is this something worth considering?

Yes, definitely. But I think the final decision should not only be driven by quantitative numbers but also on our idea of the design and previous feedback concerning this issue. Going by numbers alone, the whole world would be pitch black on pure white ;-)

alcarazr:
[...]

PR #244, opened 2016-11-21
Add metropolis-highcontrast color theme

matze:
I grew a bit tired of the default look of the theme, so here's a color theme that is even simpler and also addresses #208 and #224. Concerning the documentation, I am not sure if that should be documented the same way as the base color theme as it essentially just sets a few values.


benjamin-weiss:
First of all, I think it is a good idea to provide a minimal (rename to sth. like high contrast?) color theme. But I'm not quite sure if the colors you have chosen are ideal yet. First of all I don't get why you are using `black!98` and not simply `black` do you think that't noticeable when it is projected? I think it just slightly reduces the contrast with no benefit. Second, the `black!50` fails the WCAG test if someone uses `block=fill`. And third, to me the alert color is too "pinkish". That's probably very subjective, but to me something darker like `#AA1111` looks better and even has an AAA rating.

Ok, out of the 4 colors you used I criticized 3 of them. That sounds bad... :worried: But what do I know, I'm not a designer. Speaking of which, I think we all have mainly a scientific/engineering background, maybe we can get an actual designer on board to propose a/some color themes.

matze:
> Ok, out of the 4 colors you used I criticized 3 of them. That sounds bad... 

Not at all, and that's why it's a PR after all. `black!98` was motivated the same way `black!2` was for the background but I have no problem changing that to `black`. Same goes for `black!50`. Concerning the alert color though I think your proposal is not alerting enough. Yes, my proposal does not get triple A but I think AA is already much better than the default orange. I will incorporate your suggestions tomorrow morning.

> Speaking of which, I think we all have mainly a scientific/engineering background, maybe we can get an actual designer on board to propose a/some color themes.

Love that idea but lacking connections … anyway this PR and the idea behind was actually to get things moving concerning suggestions for a variety of color themes, even some that are not necessarily the best in terms of contrast – then b/w would be the choice in any case anyway.

matze:
Okay, alert and example on regular white backgrounds get AAA rating however the filled block option is difficult because it is *that* gray. The only sensible solution I see would be muting the color of the blocks but that would also break the look of the normal theme.

benjamin-weiss:
I think the color theme now looks nice. Especially if you use a thicker font it imho provides a quite good readability. Testing this new color theme with a thicker font i realized that it really makes sense in this case to increase the line width of the progress bar. Maybe we should extend our font customization example for thicker fonts with a tip on how to set the line width and add a section where we list all the internal lengths we introduced.

To make it easier for other to participate in the discussion here is a preview of this color theme. One the left with the regular font and on the right with a bolder font and a thicker line.

preview

Bisaloo:
I think it looks great with increased weight! Great job both of you!

matze:
:+1: @benjamin-weiss 

And yes, I even noticed with the default font weight that the section progressbar can be hard to see and am in favor of increasing it and ideally make it font width dependent if possible. One question concerning the screenshot: is it bold or semibold for the titles and blocks?

benjamin-weiss:
> And yes, I even noticed with the default font weight that the section progressbar can be hard to see and am in favor of increasing it and ideally make it font width dependent if possible.

Yes I also think increasing the line width is a good idea. I was also thinking about making it font thickness dependent, but could not come up with a good solution how to do that. But you are right, that would be ideal. Can we get some kind of information about the font thickness from `fontspec`?

> One question concerning the screenshot: is it bold or semibold for the titles and blocks?

These are the settings I used:
```TeX
\setsansfont[BoldFont={Fira Sans SemiBold}]{Fira Sans Book}
\makeatletter
\setlength{\metropolis@titleseparator@linewidth}{1pt}
\setlength{\metropolis@progressonsectionpage@linewidth}{1pt}
\setlength{\metropolis@progressinheadfoot@linewidth}{1pt}
\makeatother
```

matze:
I see, remembered it a tad lighter but still much more legible especially in badly lit rooms.

I suppose we cannot get that information from fontspec, just approximating from the font file name ... but that opens a whole other can of worms.

benjamin-weiss:
Ok, I thought about it again. What do you think about this idea? We approximate the thickness by measuring the height of a hyphen:

```TeX
\documentclass{article}
\usepackage{tikz}
\usepackage{fontspec}
\usepackage{calc}
\newlength{\myLinewidth}
\newfontfamily{\firasans}{Fira Sans}
\newfontfamily{\firasansbold}{Fira Sans Bold}
\parindent0pt

\begin{document}

\settoheight{\myLinewidth}{\firasans-}
% XXX: find a better way to normalize
\addtolength{\myLinewidth}{-3.2pt}
\setlength{\myLinewidth}{3\myLinewidth}
{\firasans Linewidth: \the\myLinewidth}

\begin{tikzpicture}
  \fill (0,0) rectangle (\textwidth, \myLinewidth);
\end{tikzpicture}\\

\settoheight{\myLinewidth}{\firasansbold-}
% XXX: find a better way to normalize
\addtolength{\myLinewidth}{-3.2pt}
\setlength{\myLinewidth}{3\myLinewidth}
{\firasans Linewidth: \the\myLinewidth}

\begin{tikzpicture}
  \fill (0,0) rectangle (\textwidth, \myLinewidth);
\end{tikzpicture}\par

\end{document}
```
This is just a prove of concept by now, but I think we can come up with a proper solution this way. The biggest problem right now is, that `\settoheight` measures the height up from the baseline, so I needed to normalize (by a magic number).

variable_linewidth

benjamin-weiss:
Using a dot I think we can get rid of the normalization:

```TeX
\documentclass{article}
\usepackage{tikz}
\usepackage{fontspec}
\usepackage{calc}
\newlength{\myLinewidth}
\newfontfamily{\firasans}{Fira Sans}
\newfontfamily{\firasansbold}{Fira Sans Bold}
\parindent0pt

\begin{document}

\settoheight{\myLinewidth}{\firasans.}
\setlength{\myLinewidth}{4\myLinewidth}
{\firasans Linewidth: \the\myLinewidth}

\begin{tikzpicture}
  \fill (0,0) rectangle (\textwidth, \myLinewidth);
\end{tikzpicture}\\

\settoheight{\myLinewidth}{\firasansbold.}
\setlength{\myLinewidth}{4\myLinewidth}
{\firasansbold Linewidth: \the\myLinewidth}

\begin{tikzpicture}
  \fill (0,0) rectangle (\textwidth, \myLinewidth);
\end{tikzpicture}\par

\end{document}
```

matze:
Wow, very creative :+1:! If you could test this with a couple of fonts (at least lmodern comes to my mind) you could open a new PR covering this change.

matze:
I will merge the color theme change now. But we should keep up with a thicker font theme and progress bar in a separate PR:

BlueRayONE:
I can't follow your conversation completely: Is the highcontrast theme now ready for usage with the current source? If yes, with which settings?

matze:
Yes, it is ready and can be enabled with regular Beamer infrastructure:
```latex
\usetheme{metropolis}
\usecolortheme{metropolis-highcontrast}
```

PR #249, opened 2017-01-06
Add missing arguments for \PackageError

josephwright:
The command needs three arguments: \@ehc is the 'Press Enter to
continue" generic advice.


josephwright:
One I might guess argue here that you could use `\PackageWarning` and not bother about the error handler.

matze:
This is superseded by #250, I suppose?

josephwright:
Yes: I probably should have done #250 from your master not from my other fix ...

Issue #247, opened 2017-01-04
metropolis does not work with beamer-3.37


kshramt:
Following code works fine with Beamer version 3.36, but does not with version 3.37.

Test code:

```latex
\documentclass{beamer}
\usetheme{metropolis}

\begin{document}

\begin{frame}
  test
\end{frame}

\end{document}
```

Error message:

```
! Package beamerfontthememetropolis Error: Patching section title failed.

See the beamerfontthememetropolis package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.263 \patchcmd
             {\beamer@@frametitle}
```

Full log:

```
This is LuaTeX, Version 0.95.0 (TeX Live 2016) 
 restricted system commands enabled.
(./slide.tex
LaTeX2e <2016/03/31> patch level 3
Babel <3.9r> and hyphenation patterns for 1 language(s) loaded.
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/latex/beamer/beamer.cls
Document Class: beamer 2017/01/01 v3.37 A class for typesetting presentations
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/latex/etoolbox/etoolbox.sty)
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/latex/beamer/beamerbasemodes.st
y
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/latex/beamer/beamerbasedecode.s
ty)) (/Users/t/usr/local/texlive/2016/texmf-dist/tex/generic/oberdiek/ifpdf.sty
)
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/latex/beamer/beamerbaseoptions.
sty (/Users/t/usr/local/texlive/2016/texmf-dist/tex/latex/graphics/keyval.sty))
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/latex/geometry/geometry.sty
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/generic/oberdiek/ifvtex.sty)
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/generic/ifxetex/ifxetex.sty))
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/latex/base/size11.clo)
(/Users/t/usr/local/texlive/2016/texmf-dist/tex/latex/pgf/basiclayer/pgfcore.st
[...]

PR #250, opened 2017-01-06
Update patch: track beamer changes

josephwright:
As raised in http://tex.stackexchange.com/questions/347383/beamer-metropolis-patchcmd-error-after-texlive-2016-update, there is a need to update the theme to track beamer changes.

### DIFF
--- a/source/beamerfontthememetropolis.dtx
+++ b/source/beamerfontthememetropolis.dtx
@@ -367,18 +367,38 @@
   {\def\insertsectionhead{\metropolis@sectiontitleformat{#2}}}
   {}
   {\PackageError{beamerfontthememetropolis}{Patching section title failed}\@ehc}
+\@tempswafalse
 \patchcmd{\beamer@section}
   {\def\insertsectionhead{\hyperlink{Navigation\the\c@page}{#1}}}
   {\def\insertsectionhead{\hyperlink{Navigation\the\c@page}{%
     \metropolis@sectiontitleformat{#1}}}}
+  {\@tempswatrue}
   {}
-  {\PackageError{beamerfontthememetropolis}{Patching section title failed}\@ehc}
+\patchcmd{\beamer@section}
+  {\protected@edef\insertsectionhead{\noexpand\hyperlink{Navigation\the\c@page}{#1}}}
+  {\protected@edef\insertsectionhead{\noexpand\hyperlink{Navigation\the\c@page}{%
+    \noexpand\metropolis@sectiontitleformat{#1}}}}
+  {\@tempswatrue}
+  {}
+\if@tempswa\else
+  \PackageError{beamerfontthememetropolis}{Patching section title failed}\@ehc
+\fi
+\@tempswafalse
 \patchcmd{\beamer@subsection}
   {\def\insertsubsectionhead{\hyperlink{Navigation\the\c@page}{#1}}}
   {\def\insertsubsectionhead{\hyperlink{Navigation\the\c@page}{%
     \metropolis@sectiontitleformat{#1}}}}
+  {\@tempswatrue}
   {}
-  {\PackageError{beamerfontthememetropolis}{Patching section title failed}\@ehc}
+\patchcmd{\beamer@subsection}
+  {\protected@edef\insertsubsectionhead{\noexpand\hyperlink{Navigation\the\c@page}{#1}}}
+  {\protected@edef\insertsubsectionhead{\noexpand\hyperlink{Navigation\the\c@page}{%
+    \noexpand\metropolis@sectiontitleformat{#1}}}}
+  {\@tempswatrue}
+  {}
+\if@tempswa\else
+  \PackageError{beamerfontthememetropolis}{Patching section title failed}\@ehc
+\fi
 %    \end{macrocode}
 %
 % Similarly, to make the |\MakeLowercase| and |\MakeUppercase| macros work in
--- a/source/beamerfontthememetropolis.dtx
+++ b/source/beamerfontthememetropolis.dtx
@@ -356,29 +356,29 @@
   {\def\inserttitle{#2}}%
   {\def\inserttitle{\metropolis@titleformat{#2}}}%
   {}%
-  {\PackageError{beamerfontthememetropolis}{Patching title failed}}
+  {\PackageError{beamerfontthememetropolis}{Patching title failed}\@ehc}
 \patchcmd{\beamer@subtitle}%
   {\def\insertsubtitle{#2}}%
   {\def\insertsubtitle{\metropolis@subtitleformat{#2}}}%
   {}%
-  {\PackageError{beamerfontthememetropolis}{Patching subtitle failed}}
+  {\PackageError{beamerfontthememetropolis}{Patching subtitle failed}\@ehc}
 \patchcmd{\sectionentry}
   {\def\insertsectionhead{#2}}
   {\def\insertsectionhead{\metropolis@sectiontitleformat{#2}}}
   {}
-  {\PackageError{beamerfontthememetropolis}{Patching section title failed}}
+  {\PackageError{beamerfontthememetropolis}{Patching section title failed}\@ehc}
 \patchcmd{\beamer@section}
   {\def\insertsectionhead{\hyperlink{Navigation\the\c@page}{#1}}}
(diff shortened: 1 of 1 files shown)


matze:
Thanks Joseph, doesn't break and I trust it will fix #247 :+1:

josephwright:
BTW,, it will fail for v3.37 as there's a bug in that release which means I've got an `\edef` not `\protected@def`. As that will only be 'in the wild' for about 5 days I hope that's acceptable!

matze:
Definitely is acceptable!

immothep:
I do not got the state about this patch. 

I got the same issue as described from @josephwright. Afterwards i've applied the patch yesterday to my Tex Installation. I've executed ```Make sty```

Before i removed the official distribution with tlmgr. 
```
sudo tlmgr remove beamertheme-metropolis
```

I've still the issue wenn creating my Document. 

```
! Package beamerfontthememetropolis Error: Patching section title failed.

See the beamerfontthememetropolis package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.291 ...opolis}{Patching section title failed}\@ehc
                                                  
```
Sould it this Error be fixed in Master? Sould i investigate to the error for further Details to this Patch?

josephwright:
@immothep  Which version of `beamer` do you have? As I've commented, v3.37 has a bug which means one of the patch lines won't apply, but that will only affect a very small number of people for a very short time (those with v3.37 of `beamer` and a development version of the theme): v3.38 is already on CTAN and has appeared in TeX Live today.

josephwright:
(I could add another patch line to cover v3.37, but really it will be a matter of a handful of people will ever see this.)

immothep:
@josephwright I was one of that handful people that used v3.37.😁  I had just installed Tex yesterday. 
Now I've upgraded my tex installation and it works.

Version of Beamer is v3.38.

```
$ tlmgr info beamer                                                                                                                         -- INSERT --
package:     beamer
category:    Package
shortdesc:   A LaTeX class for producing presentations and slides
longdesc:    The beamer LaTeX class can be used for producing slides. The class works in both PostScript and direct PDF output modes, using the pgf graphics system for visual effects. Content is created in the frame environment, and each frame can be made up of a number of slides using a simple notation for specifying material to appear on each slide within a frame. Short versions of title, authors, institute can also be specified as optional parameters. Whole frame graphics are supported by plain frames. The class supports figure and table environments, transparency effects, varying slide transitions and animations. Beamer also provides compatibility with other packages like prosper. The package now incorporates the functionality of the former translator package, which is used for customising the package for use in other language environments. Beamer depends on the following other packages: atbegshi, etoolbox, hyperref, ifpdf, and pgf.
installed:   Yes
revision:    42886
sizes:       doc: 4957k, run: 1373k
relocatable: No
cat-version: 3.38
cat-date:    2017-01-06 05:43:03 +0100
cat-license: lppl1.3gpl2
cat-topics:  presentation class
collection:  collection-latexrecommended
```
Thank you so much Joseph.

josephwright:
@immothep  Oh, I expect quite a few people will get v3.37, for a few  days: what I don't expect is that many will also have the development version of the `metropolis` theme _and_ will not update to v3.38 within a few days.

benjamin-weiss:
> what I don't expect is that many will also have the development version of the metropolis theme and will not update to v3.38 within a few days.

I think so too. So we should probably release a new version of `metropolis` on CTAN soon.

@matze What do you think which still open issues should be resolved before releasing a new version? Should we open a new issue to discuss this?

matze:
I opened #252 to track the remaining things, please comment, if you see necessary additions or issues that could be postponed.

alcarazr:
I'm using the (latest) CTAN versions of both `metropolis` (v1.1) and `beamer` (v3.38). The default test slide:
```
%!TEX TS-program = xelatex
\documentclass{beamer}
\usetheme{metropolis}

\begin{document}

\begin{frame}
  test
\end{frame}

\end{document}
```
does not compile for me (log file attached).
[test_log.txt](https://github.com/matze/mtheme/files/692934/test_log.txt).

Is there anything I can do before `metropolis` gets updated in CTAN? Will this happen "soon"?
Thanks in advance!

matze:
Don't worry, I won't push a new version to CTAN before Beamer compatibility is verified.

alcarazr:
> Don't worry, I won't push a new version to CTAN before Beamer compatibility is verified.

I appreciate that @matze, but my main problem right now is that I cannot compile previous working documents. Should I wait for `metropolis` to be updated in CTAN or manually download the latest version from GitHub for the time being?

Thanks!

matze:
Ah I see. Well either downgrade Beamer (if that is possible with your package manager) or indeed download the latest version from GitHub. I cannot give an exact release date yet but it shouldn't take too long.

alcarazr:
Thanks @matze. I was able to downgrade Beamer (to v3.36) and now I can compile again my documents. Was this sudden change in Beamer expected/announced?

josephwright:
@alcarazr  The change in `beamer` is to an internal definition: the fact that it's patched by the `metroplois` theme perhaps shows an additional interface would be useful but it's not possible to say 'nothing will ever change' in `beamer` unless all fixes and improvements are entirely halted. It's slightly unfortunate that the patch set up had an issue (meaning that simply saying 'continue' isn't entirely reliable) and as I've commented I wonder if a warning rather than an error might be more appropriate.

bluddy:
This seems to still be a problem when installing metropolis using `tlmgr`. What's the easiest way to resolve it?

josephwright:
@bluddy The version on CTAN (and thus in TeX Live) is still v1.1: you'll have to wait for a release to update using `tlmgr`.

alcarazr:
@bluddy The easiest way to resolve it using the version on CTAN is not updating beamer (leave it at v3.36, downgrade if you automatically updated and are able to do it). @matze Perhaps an updated version (not v2.2) could be uploaded to CTAN, just fixing the beamer-related bug, so that users can safely update their distributions using tlmgr. I like having my TeX distribution updated, but right now I have to select all packages but `beamer` when updating to prevent `metropolis` to stop working.

bluddy:
Just to give an idea, metropolis is part of my workflow of creating class slides from markdown using pandoc. I need to give other instructors access to this workflow, and the easier it is to use, the better -- otherwise it defeats the purpose of using pandoc and markdown for ease of use in the first place.

Unfortunately it doesn't seem like `tlmgr` supports downgrading of packages, which means I  need to dive in and hack it - which I don't want to do - I want to produce content, not struggle with tooling.

josephwright:
@bluddy `tlmgr` has had a backup option for some time: `tlmgr restore beamer` will list those available. (I agree that a release of the theme here would be best, but that's down to the maintainer.)

ddeimeke:
Add the working style files to the same directories as your source files are. Local style files are considered first. That would be the easiest way to cope with the situation.

bluddy:
@ddeimeke The style files were included before, but they didn't work with the new beamer. I was hoping to get past the fragility of including them, but it seems like the best option now.

@josephwright thanks for letting me know, but that's still not as good as a downgrade option. I have to have the right version installed locally, which I don't.

josephwright:
@bluddy 'Downgrade' is not supported as CTAN and the TeX Live mirrors have only the latest version of material available: `backup` relies on a local copy. That said, I think I can 'fix' things from the `beamer` end using a rather ugly hack: I can't prevent the (broken) patching but I can make sure it finds something to patch. I'll add a no-op piece of code to `beamer` which should deal with the error, though as I say I'd hope to see the theme updated soon.

josephwright:
@bluddy I've sent https://github.com/josephwright/beamer/releases/tag/v3.40 to CTAN: thsi should work around the issue until the theme is updated

alcarazr:
Thanks a lot @josephwright, really appreciate how you care not only about `beamer`, but also about `metropolis`.

bluddy:
Thanks so much @josephwright.

Sounds like, in general, CTAN should be using a git-based workflow. Obviously it came about before git did, but it would provide many benefits, including downgrades.

matze:
Alright, I will push an intermediate bugfix version that will handle this situation from our side as well. I am getting numerous emails complaining about the broken state. @josephwright I'll let you know when it's uploaded.

PR #254, opened 2017-01-10
Format table rules following booktabs advice

josephwright:


matze:
I learn something new every day :+1:

PR #304, opened 2017-12-02
Patch latest beamer section commands

### LINKED ISSUE #303: Fails to compile with beamer 3.45
Metropolis fails to compile with beamer version 3.45 while 3.44 doesn't cause this issue. The problematic file should be `beamerbasesection.sty`.

Error:
```
! Package beamerfontthememetropolis Error: Patching section title failed.
```

Minimal non-working example:
```
\documentclass{beamer}
\usetheme{metropolis}
\begin{document}
\end{document}
```

The error log is attached: [main.log](https://github.com/matze/mtheme/files/1523796/main.log)

josephwright:
This fixes #303.

### DIFF
--- a/source/beamerfontthememetropolis.dtx
+++ b/source/beamerfontthememetropolis.dtx
@@ -369,6 +369,12 @@
   {\PackageError{beamerfontthememetropolis}{Patching section title failed}\@ehc}
 \@tempswafalse
 \patchcmd{\beamer@section}
+  {\edef\insertsectionhead{\noexpand\hyperlink{Navigation\the\c@page}{\unexpanded{#1}}}}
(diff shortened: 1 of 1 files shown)


matze:
> one might argue that a failure to update design in a 'consistent' way is an error

One might also argue that interfaces should be as stable as possible and pieces lower in the stack strive to avoid breaking things higher up ;-)

josephwright:
@matze I meant updating (on a user PC) one of `beamer` or `metropolis` but not the other.

BTW, this is all patching internals: the issue really is that we don't have a 'make things X case' interface I guess.

josephwright:
@matze  Ah, I re-read the comment you are quoting: what I was getting at is that it's your call that the path failing is an error not a warning (it's an important part of the design).

PR #325, opened 2018-05-10
Updated Fira font paths

jzaremba:
Fira fonts have moved to https://bboxtype.com/typefaces/FiraSans/ and https://github.com/bBoxType/FiraSans.
This updates the getFiraFont.sh script to download the files from the new location.
Also removed calls to 'sudo', which is not installed and is unnecessary for the docker image.
Fixed docker build command in Makefile.


matze:
Thanks :+1:

Issue #335, opened 2018-07-22
Standout option affects all following slides with ignorenonframetext


pacien:
_This issue is related to but distinct from #334._

The `standout` option wrongly affects all the following slides when using the `ignorenonframetext` class option.

~~This might probably be solved by implementing this: https://tex.stackexchange.com/a/181042.~~

## Example

![standout-ignorenonframetext](https://user-images.githubusercontent.com/1449319/43047546-f7f6d356-8dd8-11e8-9e6a-c83df648953e.png)

```latex
\documentclass[ignorenonframetext]{beamer}
\usetheme{metropolis}
\begin{document}

  \begin{frame}{Normal frame}
  \end{frame}

  \begin{frame}[standout]
    Standout frame
  \end{frame}

  \begin{frame}{Normal frame}
  \end{frame}

\end{document}
```

basbebe:
This still seems to be a problem for me when using pandoc

pacien:
Quoting basbebe (2021-11-10 09:36:09)
> This still seems to be a problem for me when using pandoc

Are you using the theme from the master branch? The fixes have been
merged there but haven't been published in a release yet.

basbebe:
I'm sorry, this is apparently what happened:

I did `make sty` in master and copied the resulting `.sty` files to `beamertheme/` in my pandoc project.\
I added `theme: metropolis` in my `default.yaml`.\
Then I got the described behavior.

When I copied the files to my project root, the problem was gone.

I deleted `theme: metropolis` in my `default.yaml` and added to my `settings.tex`:
```latex
\usepackage{beamertheme/beamercolorthememetropolis}
\usepackage{beamertheme/beamerfontthememetropolis}
\usepackage{beamertheme/beamerinnerthememetropolis}
\usepackage{beamertheme/beamerouterthememetropolis}
\usepackage{beamertheme/beamerthememetropolis}
```

had to omit `beamertheme/beamercolorthememetropolis-highcontrast.sty` (contrast too high)
 and `beamertheme/pgfplotsthemetol.sty` (error when running `pandoc`)

PR #336, opened 2018-07-22
Restore theme within frame, making standout slides work with ignorenonframetext

pacien:
Should solve issue #335.

State is well restored even with `ignorenonframetext`.

~~Notes are still impacted though.~~ Not anymore, see PR #337.

standout-ignorenonframetext-fixup


matze:
Could you reformulate the commit message (limit yourself to 50 characters in the summary line, add as much as you need in the body) and prepend it with `Fix #335`?

pacien:
Sure, it's done.

matze:
Thanks :+1:

Issue #334, opened 2018-07-22
Standout slides theme affects notes


pacien:
The `standout` frame option affects nested note blocks, breaking their layout: the date and all the content is centered whereas it shouldn't.

While it is possible to circumvent this issue by putting the `\note` block outside of the frame block, the Beamer User Guide states that nesting it inside is recommended. It is also not possible to put it outside when using Pandoc.

## Example

![standout-notes](https://user-images.githubusercontent.com/1449319/43047398-119b59a0-8dd7-11e8-89e6-2e31fa82f6f2.png)

```latex
\documentclass{beamer}
\usetheme{metropolis}
\usepackage{pgfpages}
\setbeameroption{show notes on second screen=right}
\begin{document}

  \begin{frame}[standout]
    Standout frame
    \note{Some notes}
  \end{frame}

\end{document}
```

PR #337, opened 2018-07-22
Restrict standout font and alignment to inner content

pacien:
Solves issue #334.

standout-notes-fixup


matze:
I will merge if you reformulate the commit message like in #336.

pacien:
Done.

Issue #274, opened 2017-04-15
Progress bars should max out at 100%


kno10:
When e.g. stopping the "total" page count at the last slide, sections in backup slides (which then will have frame numbers like 20/15) will have their progress bar go beyond 100%.

benjamin-weiss:
Sorry, but I cannot understand the issue here. As stated in the [documentation](http://mirrors.ctan.org/macros/latex/contrib/beamer-contrib/themes/metropolis/doc/metropolistheme.pdf) in section 5.1 you should use the `appendixnumberbeamer` package. If you do so, the progress bar and page numbering  is automatically deactivated for backup slides.

kno10:
I don't use that package, because I don't want the numbering to start with 0 again.

I do, however, call `\appendix`. The progress bar disappears on the outer theme, but it remains on the `sectionpage` slides, where it is most noticeable that they exceed 100%.

benjamin-weiss:
Ok. Can you put together a MWE?

kno10:
I would also like the appendix frames to have numbers, too. I used to number them simply 26/25 etc. if the main part was 25 frames. Then people can still refer to them as "slide 26". Thus, I don't think metropolis should automatically turn of page numbering in the appendix.
```
\documentclass{beamer}
\usetheme{metropolis}
\begin{document}

\frame{Main}

% Store the real frame count.
\newcounter{framenumberBeforeAppendix}
\setcounter{framenumberBeforeAppendix}{\value{framenumber}}

% Actually I would like to have page numbers here, too. 
\appendix

\frame{Appendix}
\frame{Appendix 2}

\section{Appendix: Progress bar is broken}

% Restore the real frame count.
\addtocounter{framenumberBeforeAppendix}{-\value{framenumber}}
\addtocounter{framenumber}{\value{framenumberBeforeAppendix}} 
\end{document}
```
I use own macros `\lastslide` and `\restorepagecount` that do the frame number magic, but it is just 2 lines each, so I inlined them here for the MWE.

PR #343, opened 2018-09-13
 Use `pgfmathsetlength`, and max progress bars out at 1

kno10:
Use `pgfmathsetlength`, and max progress bars out at 1.
This should also avoid the problem of a too long line on the first compile.


kno10:
This should resolve #274; but this pull may still have a syntax error - I edited this in github. Also you may want to squash the two commits when merging.

